PR #62, opened 2018-01-17
Execute seeds only in dev environment.

rhnonose:


priv/repo/seeds.exs
```diff
@@ -1,206 +1,198 @@
...
-})
-
-{:ok, interest4} = Repo.insert(%Interest{
-  name: "Interested Person 3",
-  email: "interested3@email.com",
-  phone: "432112344321",
-  message: "Looks like an awesome listing",
-  listing: listing4
-})
+if Mix.env == :dev do
```
gusaiani: 👍era isso, obrigado

PR #99, opened 2018-02-01
Add filter listings for max/min latitude/longitude.

rhnonose:


test/re/listings/listings_test.exs
```diff
@@ -10,9 +10,13 @@ defmodule Re.ListingsTest do
 
   describe "all/1" do
     test "should filter by attributes" do
-      laranjeiras = insert(:address, street: "astreet", neighborhood: "Laranjeiras")
-      leblon = insert(:address, street: "anotherstreet", neighborhood: "Leblon")
-      botafogo = insert(:address, street: "onemorestreet", neighborhood: "Botafogo")
+      laranjeiras =
+        insert(:address, street: "astreet", neighborhood: "Laranjeiras", lat: -1, lng: -1)
```
gusaiani: @rhnonose Only as a precaution, perhaps we may use coords in the actual GMaps float precision levels?

Ex:
https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/float.ex#L161

rhnonose: Sure thing.

PR #105, opened 2018-02-08
Fix user public api

rhnonose:
- Refactor router to:
  - put `EnsureAuthenticated` individually to controllers/actions
  - use the rest of the pipeline for public apis as well

### DIFF
--- a/lib/re_web/controllers/listing_controller.ex
+++ b/lib/re_web/controllers/listing_controller.ex
@@ -4,7 +4,6 @@ defmodule ReWeb.ListingController do
 
   alias Re.{
     Addresses,
-    Images,
     Listings,
     Stats.Visualizations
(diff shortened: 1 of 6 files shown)


lib/re_web/controllers/listing_controller.ex
```diff
@@ -13,6 +12,17 @@ defmodule ReWeb.ListingController do
 
   action_fallback(ReWeb.FallbackController)
 
+  plug(
+    Guardian.Plug.EnsureAuthenticated
+    when action in [
+           :create,
```
gusaiani: @rhnonose Tabs here?

rhnonose: I don't know, the formatter set like that.

PR #114, opened 2018-02-20
Relax neighborhood filter.

rhnonose:
- Use arbitrary mapping to sort out close neighborhood
- Add one close neighborhood per neighborhood on array, removing duplicated ones

### DIFF
--- a/lib/re/addresses/neighborhoods.ex
+++ b/lib/re/addresses/neighborhoods.ex
@@ -22,4 +22,20 @@ defmodule Re.Neighborhoods do
   def all do
     Repo.all(@all_query)
   end
+
+  @doc """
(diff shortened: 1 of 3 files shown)


gusaiani:
@rhnonose I'd rather leave this branch around, without merging yet.

If, by the time we use the relaxed filters in the frontend, you didn't do this via expanding the geolocation "rectangle", we merge this one.

It's a pretty clever idea for something very quick though.

What do you think?

rhnonose:
@gusaiani I'd prefer merging it to test the "concept" and, since the API itself doesn't have to change, only the "origin" of the data, we'll not be breaking anything when we exchange for a better approximation.

PR #123, opened 2018-03-07
Put is_active in v2 listing insertion/update.

rhnonose:


lib/re/listings/listing.ex
```diff
@@ -55,7 +55,7 @@ defmodule Re.Listing do
 
   @attributes_v2 ~w(type complement description price property_tax maintenance_fee floor
```
gusaiani: @rhnonose Is the `_v2` related to our update in the frontend?

If so, it is ok for us to have the latest version be the only version in the backend as well?

rhnonose: Yes, it is.
After some time and things look fine, I'll remove the old code.

PR #141, opened 2018-03-21
Add route to activate listing

rhnonose:


gusaiani:
Ebert has finished reviewing this Pull Request and has found:

* 1 fixed issue! 🎉

You can see more details about this review at https://ebertapp.io/github/em-casa/backend/pulls/141.

lib/re_web/controllers/listing_controller.ex
```diff
@@ -85,6 +85,14 @@ defmodule ReWeb.ListingController do
          do: send_resp(conn, :no_content, "")
   end
 
+  def activate(conn, %{"id" => id}, user) do
```
gusaiani: @rhnonose Wouldn't this be more restful if in a nested controller, like `listings/active:update`?

I'm very open to this being a first attempt with GraphQL too.

rhnonose: The GraphQL suggestion is great, I'll consider.

The nested controller is tricky. I don't think in any way we'll "conform" to standards with an RPC style naming since the action is outside of the RESTful nomenclature, so the nested part sounds a little bit unnecessary.

rhnonose:
I'll setup a GraphQL endpoint to start off with this.

PR #146, opened 2018-03-22
Add activate listing as graphql endpoint

rhnonose:
- Setup absinthe for GraphQL endpoint
- Add `activateListing` mutation


test/re_web/graphql/listings/listings_test.exs
```diff
@@ -0,0 +1,74 @@
+defmodule ReWeb.GraphQLTest do
```
gusaiani: @rhnonose 
Perhaps should we call this `ReWeb.GraphQL.ListingsTest`?

PR #185, opened 2018-04-26
Notify address update

rhnonose:
- Refactor address structure (use `Repo.insert_or_update/1`)
- Notify admin if the address is updated

### DIFF
--- a/lib/re/addresses/addresses.ex
+++ b/lib/re/addresses/addresses.ex
@@ -21,22 +21,25 @@ defmodule Re.Addresses do
   end
 
   def insert_or_update(params) do
-    changeset = case get(params) do
-      {:error, :not_found} -> build_address(params)
-      {:ok, address} -> address
-    end
-    |> Address.changeset(params)
+    changeset =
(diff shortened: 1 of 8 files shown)


lib/re/addresses/addresses.ex
```diff
@@ -8,49 +8,35 @@ defmodule Re.Addresses do
...
+  def insert_or_update(params) do
+    changeset = case get(params) do
+      {:error, :not_found} -> build_address(params)
+      {:ok, address} -> address
     end
-  end
-
-  defp changed?(address, address_params) do
-    %{changes: changes} = Address.changeset(address, address_params)
+    |> Address.changeset(params)
```
gusaiani: Use a function call when a pipeline is only one function long

gusaiani:
Ebert has finished reviewing this Pull Request and has found:

* 1 possible new issue (including those that may have been commented here).

You can see more details about this review at https://ebertapp.io/github/em-casa/backend/pulls/185.

PR #227, opened 2018-06-13
Finish listings graphql query

rhnonose:
- Load inactive images only for admins and owners
- Resolve `owner` field only for admins and owners


lib/re/listings/listings.ex
```diff
@@ -15,7 +15,11 @@ defmodule Re.Listings do
 
   alias Ecto.Changeset
 
-  defdelegate authorize(action, user, params), to: Re.Listings.Policy
+  defdelegate authorize(action, user, params), to: __MODULE__.Policy
+
+  # def data(params), do: Dataloader.Ecto.new(Repo, query: &query/2, default_params: params)
```
gusaiani: @rhnonose, perhaps we should delete this commented line and #22 as well?

rhnonose: Aw thanks, forgot that one.

PR #380, opened 2019-01-10
Zap integration slug

pmargreff:
The main idea in this PR is to require Zap's exportation path to use slugs (i.e.: `state_slug/city_slug`) for fetching only listings located on respective city and state. It also allows params as `page_size` and `offset` for better integration with our platform requirements.


pmargreff:
Feel free to point all possible bugs or improvements which I didn't catch. Also, should us handle invalid paths; paths without slugs (like the original one), or with only one (i.e.: `exports/zap/rj`) and show these path errors to the end users?

I'll open another PR for the highlights since I guess it has some complexity, this approach looks ok to you? @rhnonose

apps/re/lib/listings/listings.ex
```diff
@@ -204,4 +204,14 @@ defmodule Re.Listings do
     |> Queries.preload_relations(@full_preload)
     |> Repo.all()
   end
+
+  def exportable(%{state_slug: state_slug, city_slug: city_slug}, params) do
+    Queries.active()
+    |> Queries.by_state_slug_and_city_slug(state_slug, city_slug)
```
rhnonose: You can use the `Re.Filtering` module to build this query.

apps/re/lib/listings/listings.ex
```diff
@@ -204,4 +204,14 @@ defmodule Re.Listings do
     |> Queries.preload_relations(@full_preload)
     |> Repo.all()
   end
+
```
rhnonose: Not sure if this function should be here in this module since the filtering/order logic belongs to the exporting functionality.

apps/re/test/listings/listings_test.exs
```diff
@@ -408,6 +408,79 @@ defmodule Re.ListingsTest do
     end
   end
 
+  describe "exportable/1" do
```
rhnonose: These tests should be in its own file (if you agree to extract the previous mentioned function out of `Re.Listings`)

apps/re/lib/listings/listings.ex
pmargreff: I have no sure either, I'll extract (probably to a `Re.Listings.Exports` module). If you think this function has a better fit in another place, feel free to suggest.

rhnonose: Sounds good.

pmargreff:
> ... it should return the proper error code (check in the integration documentation if there's a specific error code to return that is specified by them)

Ok, I'll take a look at that.

pmargreff:
I didn't find any error instruction in the documentation. I introduced a brief error message with an appropriate status code. Can you review this PR again @rhnonose ?

PR #384, opened 2019-01-22
Deactivate images

rhnonose:
- `imageDeactivate` mutation
  - Receives a list of images, checks for `listing_id` in common
  - Returns the list of images deactivated
  - Admin and user that owns the listing
- `imageDeactivated` subscription
  - Receives a list of images that were deactivated
  - Admin-only subscription (at least for now)

### DIFF
--- a/apps/re/lib/images/images.ex
+++ b/apps/re/lib/images/images.ex
@@ -77,6 +77,12 @@ defmodule Re.Images do
     end
   end
 
+  def list_by_ids(ids) do
+    ids
+    |> Queries.with_ids()
+    |> Repo.all()
+  end
+
   defp do_get(input) do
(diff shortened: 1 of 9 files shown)


apps/re/lib/images/images.ex
```diff
@@ -91,6 +97,13 @@ defmodule Re.Images do
     end
   end
 
+  def is_same_listing(images) do
```
pmargreff: The name made me understand it would return a `boolean`, I guess it's mainly because the `is_` prefix, does it make sense for you?

Possible names: `fetch_listing`, `get_listing`, `get_listing_from_image` or even `find_listing_from_images`. These are some suggestions, but feel free to ignore them in favor of another one.

PR #385, opened 2019-01-22
Remove graphql references to zap's highlights

pmargreff:
Remove only Zap's highlights related references on graphql.

### DIFF
--- a/apps/re_web/lib/graphql/resolvers/dashboard.ex
+++ b/apps/re_web/lib/graphql/resolvers/dashboard.ex
@@ -90,33 +90,9 @@ defmodule ReWeb.Resolvers.Dashboard do
     end
   end
 
-  def listing_zap_highlights(params, %{context: %{current_user: current_user}}) do
-    with :ok <- is_admin(current_user),
-         do: {:ok, Highlights.get_zap_highlights(params)}
-  end
-
-  def listing_zap_super_highlights(params, %{context: %{current_user: current_user}}) do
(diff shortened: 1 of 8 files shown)


pmargreff:
Not so simple. I'll reuse a lot of the original structure (XML builder and so on) on the other PR. If I remove it here I'll need to reintroduce it later (mainly the tests), we'll have a bunch of not edited code on the diff. 

The intention here is to remove only the graphql references, and not the structure at all. If it's not good enough I suggest only open one bigger PR with all changes, it looks more reasonable than open 1 PR remove a lot of code and another PR reintroducing the same code, which will result in 2 big PR at the end. 

Please, @rhnonose let me know your thoughts about these options.

rhnonose:
Hm, I reckon from the other PR that it would removed. Feel free to decide where it goes then :)

pmargreff:
> Hm, I reckon from the other PR that it would removed.

You got it totally right, I'll remove all actual module and table. But the main point is: If I remove these structures the dependent modules will break, like [tests for the `Exporter`](https://github.com/emcasa/backend/blob/master/apps/re/test/exporters/zap_test.exs#L1), so I would need to remove both tests and respective module right on this branch for reintroducing them in the other PR. 

Did you get it? Does this make sense? If not, we can chat on private and find some alternative.

rhnonose:
Yeah, that's what I meant since this is a cleanup PR.

PR #386, opened 2019-01-22
Add listing updated subscription

rhnonose:
Add `listingUpdated` subscription for admin-only (at least for now).


pmargreff:
@rhnonose I guess we have a flip test here. It will (probably) pass if you run the CI again. I created a new task for this problem on Notion.

rhnonose:
Yep, passed just fine, thanks.

PR #391, opened 2019-01-24
Reduce number of highlights for both cities

pmargreff:
While reading the task again I noticed that I used the total number of listings as the highlights size. Reduce to the right size on this PR.


rhnonose:
Shall we put this in an environment variable?

It might change rather frequently, we could reduce the necessity to change the code.

pmargreff:
Yes, makes sense. I'm changing it and I will add in this PR.

pmargreff:
@rhnonose done, please double check the variable names on env vars panel as well.

PR #395, opened 2019-01-28
Fix listing update notification

rhnonose:
- Fix wrong pattern-match in `emails/server.ex` and test `handle_info` call rather than `handle_cast` call
- Refactor activation to use `PubSub.pubish_update` with changeset
- Rename price and status history modules


apps/re_integrations/lib/notifications/emails/user.ex
```diff
@@ -88,7 +88,7 @@ defmodule ReIntegrations.Notifications.Emails.User do
                   <a href=\"#{listing_url}\">Imóvel</a>")
   end
 
-  def listing_updated(%User{name: name, email: email}, %Listing{} = listing, changes) do
+  def listing_updated(%Listing{user: %User{name: name, email: email}} = listing, changes) do
```
joaodubas: This is the only method that doesn't match a `User` and a `Listing`. Does it make sense to do this change?

rhnonose:
Usually, plurals like `Users` and `Listings` are for context modules, which is a concept sort of introduced in Phoenix 1.3 and gets some inspiration from Domain-Driven Design.

Previously, Phoenix would generate code inside REST controllers, so there was no "business-level" abstraction in its basic structure.

apps/re_integrations/lib/notifications/emails/user.ex
rhnonose: It's kind of a defensive coding and a little bit old so it's not strictly necessary. Do you think we should remove it?

joaodubas: I don't see a reason to remove it. Maybe we could add another PR, changing [`listing_added_admin`](https://github.com/emcasa/backend/blob/744cba5a46ab5f0c1536f6ff850f44f7958748e0/apps/re_integrations/lib/notifications/emails/user.ex#L74-L89) to keep the same matching.

rhnonose: Nice, you can create a task in notion with `refactoring` for that :)

PR #397, opened 2019-01-28
Do not send listing update e-mail if the modifier is an admin

rhnonose:
- Fix previously introduced e-mail fix that was sending e-mail when admin updated the listing
- Refactor `PubSub` to include metadata

### DIFF
--- a/apps/re_integrations/lib/notifications/emails/user.ex
+++ b/apps/re_integrations/lib/notifications/emails/user.ex
@@ -88,7 +88,7 @@ defmodule ReIntegrations.Notifications.Emails.User do
                   <a href=\"#{listing_url}\">Imóvel</a>")
   end
 
-  def listing_updated(%Listing{user: %User{name: name, email: email}} = listing, changes) do
+  def listing_updated(listing, user, changes) do
     listing_url = build_url(@listing_path, to_string(listing.id))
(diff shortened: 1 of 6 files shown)


apps/re_integrations/lib/notifications/emails/user.ex
```diff
@@ -88,20 +88,22 @@ defmodule ReIntegrations.Notifications.Emails.User do
                   <a href=\"#{listing_url}\">Imóvel</a>")
   end
 
-  def listing_updated(%Listing{user: %User{name: name, email: email}} = listing, changes) do
+  def listing_updated(listing, user, changes) do
```
joaodubas: Almost the same matching as the previous one.

rhnonose: Yeah, before I was getting `user` from the relationship in `listing` but now I'm passing a `metadata` when publishing to the `PubSub`, so now this signature makes sense.

rhnonose:
Yeah, I forgot to run `mix format`, we should automate it.

joaodubas:
We can add it as git hook.

rhnonose:
Yeah, a pre-commit hook would be perfect.

joaodubas:
Creating a PR with this.

PR #399, opened 2019-01-28
git hook to check format before commit

joaodubas:
To avoid situations where `mix format` isn't run before commiting, we can add this `pre-commit` hook, that checks that every staged change is properly formatted.

TODO:

- [x] create script to run before commit
- [x] add mix target to install script in `.git/hooks`

### DIFF
--- a/priv/git/pre-commit
+++ b/priv/git/pre-commit
@@ -1,11 +1,13 @@
 #!/bin/sh
 to_check=$(git diff --name-only --cached | grep '\.ex')
-if [ $(echo ${to_check} | wc -l) == 1 ]; then
+to_check_len=$(echo ${to_check} | wc -l)
+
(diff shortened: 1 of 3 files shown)


priv/git/pre-commit
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+to_check=$(git diff --name-only --cached | grep '\.ex')
```
rhnonose: The formatter also runs on `.exs` files.

joaodubas: The pattern matches both `*.ex` and `*.exs`. Example in [asciinema](https://asciinema.org/a/ZmDwaSkNBmbp5qgwXZKk9a75X).

rhnonose: My bad.

PR #400, opened 2019-01-29
Fix configs using wrong otp applicaitons

rhnonose:
Some configurations were using the main `re` application as otp app and some apps were being launched under `re` as well.
Refactor to use appropriate `re_web` and `re_integrations` applications


apps/re_integrations/lib/pipedrive/pipedrive.ex
```diff
@@ -5,7 +5,7 @@ defmodule ReIntegrations.Pipedrive do
   require Logger
 
   @activity_params ["deal_id", "marked_as_done_time"]
-  @server Application.get_env(:re, :pipedrive, __MODULE__.Server)
+  # @server Application.get_env(:re_integrations, :pipedrive, __MODULE__.Server)
```
joaodubas: Why comment this line?

rhnonose: This was sort of a "mock" to not call it but it wasn't being used (the test module in `test.exs` wasn't event declared), so I tested removing it but forgot to remove the commented code.

config/dev.secret-example.exs
```diff
@@ -22,20 +22,22 @@ config :re, ReWeb.Guardian,
     "y" => "xxxxxx"
   }
 
-# The emails below are used when a visitor
-# fills out a form expressing interest in a listing.
-# Both `to:` and `from:` can be individual emails
```
pmargreff: This comment (mainly the part about how to fill the emails vars) looks informative could be a good idea keep it (if you don't mind, of course).

config/test.exs
```diff
@@ -37,22 +33,21 @@ config :re, Re.Repo,
 config :account_kit,
   app_id: "123"
 
-config :re, ReIntegrations.Notifications.Emails.Mailer, adapter: Swoosh.Adapters.Test
-
-config :re, :visualizations, Re.TestVisualizations
-config :re, :emails, ReIntegrations.Notifications.Emails.UserTest
-config :re, :elasticsearch, ReWeb.TestSearch
-config :re_integrations, :http, ReIntegrations.TestHTTP
-config :re, :pipedrive, ReWeb.TestPipedriveServer
```
pmargreff: Didn't find this module to be excluded, but I guess it was excluded in another previous commit...

PR #401, opened 2019-01-29
Vivareal integration

pmargreff:
Main changes: 
- Enable slugs on Vivareal path.  
- Allow Vivareal's exporter to receive `highlight_ids` as an option. 
- Refactor `convert_attributes\2` to `convert_attribute/3` to support it to receive highlights list and keep as an optional parameter (will port it to zap as well).


pmargreff:
One thing I didn't solve and would appreciate your opinion @rhnonose, @joaodubas. 

[This query](https://github.com/emcasa/backend/pull/401/files#diff-efaf3993b366d4496f8f38f8a2a94139R48) is being replicated on [Zap's Plug](https://github.com/emcasa/backend/blob/master/apps/re_web/lib/exporters/zap/plug.ex#L71) and would be a good idea to share it, once we probably will need to change both when any change is needed. But I cannot identify the best place to create a function with this query. 

Can I create a function with this query in `Highlights` module or it would be considered be an anti-pattern once all queries are in `Queries` modules? If this function should be in `Queries` module, should I create a new one under `exporters` directory?

apps/re/lib/listings/highlights.ex
```diff
@@ -2,6 +2,18 @@ defmodule Re.Listings.Highlights do
...
+                                        :re,
+                                        :vivareal_highlights_size_sao_paulo,
+                                        0
+                                      )
+  @vivareal_highlights_size_rio_de_janeiro Application.get_env(
+                                             :re,
+                                             :vivareal_highlights_size_rio_de_janeiro,
+                                             0
+                                           )
+
```
joaodubas: We could separate the highlights from viva real and zap in their own modules.

apps/re_web/test/exporters/vivareal/plug_test.exs
```diff
@@ -6,10 +6,37 @@ defmodule ReWeb.Exporters.Vivareal.PlugTest do
...
+    test "should render listings XML", %{conn: conn} do
+      conn = get(conn, "/exporters/vivareal/rj/rio-de-janeiro")
 
-    assert response(conn, 200) == @empty_listings
+      assert response(conn, 200) == @empty_listings
+    end
+  end
+
+  describe "with inactive state and city slugs on path" do
+    test "should export zap XML", %{conn: conn} do
```
joaodubas: Is the name of this test correct?

apps/re/lib/listings/highlights.ex
pmargreff: Only the env vars? Are you thinking about something [like this](https://github.com/emcasa/backend/commit/d86b5bd342452ff4777e6b7f842688fc9b678186)? Tried this in zap feature, but I [reverted it later](https://github.com/emcasa/backend/commit/065c5fb2fe8c958b9aa989b2398abdda655dc4c8) because all env vars namespaces requires or an application or a dependency with the same name. So if we have something like: `zap` it will raise a warning because cannot found the associated application. 

Maybe we can solve the readability problem nesting the vars like [this one](https://github.com/emcasa/backend/blob/master/config/prod.exs#L15). What do you think about it?

apps/re_web/test/exporters/vivareal/plug_test.exs
pmargreff: Yeah, but I guess it's not meaningful enough, I will rewrite it.

apps/re/lib/listings/highlights.ex
joaodubas: The comment end up in a bad place, I think we could have something like:

* `.../listings/highlights.ex`: common highlights functionality
* `.../listings/highlights_vivareal.ex`: highlights logic specific to viva real
* `.../listings/highlights_zap.ex`: highlights logic specific to zap

joaodubas:
Maybe we could define a query in `Re.Listings.Queries` that receive all filters and implemented the same operations, so both zap and viva-real `Plug`s could use the same query.

```elixir
# file apps/re/lib/listings/queries/queries.ex
  ...
  def fetch_highlights(query, filters) do
    query
    |> active()
    |> Filtering.apply(filters)
    |> preload_relations([:address])
    |> order_by_id()
  end
  ...
```

apps/re/lib/listings/highlights.ex
rhnonose: I think this module in particular is ok to be only one since it's rather small and groups funcionalities that are common between the highlights.

rhnonose:
You should also add the new configs to `dev.secret-example.exs`.

config/dev.secret-example.exs
```diff
@@ -32,7 +32,13 @@ config :re,
...
   pipedrive_token: "your_token",
   pipedrive_webhook_user: "pipedrive_user",
-  pipedrive_webhook_pass: "pipedrive_pass"
+  pipedrive_webhook_pass: "pipedrive_pass",
+  vivareal_highlights_size_rio_de_janeiro: "VIVAREAL_HIGHLIGHTS_SIZE_RIO_DE_JANEIRO",
+  vivareal_highlights_size_sao_paulo: "VIVAREAL_HIGHLIGHTS_SIZE_SAO_PAULO",
+  zap_highlights_size_rio_de_janeiro: "ZAP_HIGHLIGHTS_SIZE_RIO_DE_JANEIRO",
+  zap_highlights_size_sao_paulo: "ZAP_HIGHLIGHTS_SIZE_SAO_PAULO",
+  zap_super_highlights_size_rio_de_janeiro: "ZAP_SUPER_HIGHLIGHTS_SIZE_RIO_DE_JANEIRO",
+  zap_super_highlights_size_sao_paulo: "ZAP_SUPER_HIGHLIGHTS_SIZE_SAO_PAULO"
```
rhnonose: Now that I'm thinking, why is this a secret? Can't it be just a regular config, with a default of 10 for dev/test?

pmargreff: Yes, it don't need to be a secret. I'll update.

apps/re/lib/listings/highlights.ex
pmargreff: > I think this module in particular is ok to be only one since it's rather small and groups funcionalities that are common between the highlights. 

Yep, and after removing the references to the old structure [it will look small](https://github.com/emcasa/backend/blob/vivareal-remove-old-structure/apps/re/lib/listings/highlights.ex).

apps/re/lib/listings/queries/queries.ex
```diff
@@ -101,4 +102,11 @@ defmodule Re.Listings.Queries do
       where: ^listing.address.city == a.city
     )
   end
+
+  def highlights(query \\ Listing, filters) do
+    active()
+    |> Filtering.apply(filters)
+    |> preload_relations([:address])
+    |> order_by_id()
+  end
```
joaodubas: Cool ❤️

PR #405, opened 2019-01-31
Truncate postal code with 5 digits

pmargreff:
Some listings would be not exported to portals due to the invalid postal code. Google maps probably return some postal codes in a 5 digits format (old and invalid format), on these cases we are now truncating "-000" (nevertheless it does not reflect the generated code be the correct one). 

While digging to this I found [we removed the validation on zip code some time ago](https://github.com/emcasa/backend/commit/114c14a4cf094adaa9255137d782a40f7c083b6f) (probably duo the same issue), should it be reinserted now all the know cases are formatted?

### DIFF
--- a/apps/re/lib/addresses/address.ex
+++ b/apps/re/lib/addresses/address.ex
@@ -47,7 +47,7 @@ defmodule Re.Address do
     |> unique_constraint(:postal_code, name: :unique_address)
     |> validate_number(:lat, greater_than: -90, less_than: 90)
     |> validate_number(:lng, greater_than: -180, less_than: 180)
-    |> truncate_postal_code()
+    |> pad_trailing_with_zeros()
     |> generate_slugs()
   end
 
@@ -58,7 +58,7 @@ defmodule Re.Address do
   end
 
 
-  def truncate_postal_code(changeset) do
+  def pad_trailing_with_zeros(changeset) do
     postal_code = Ecto.Changeset.get_field(changeset, :postal_code, "")
 
     if partial_postal_code?(postal_code) do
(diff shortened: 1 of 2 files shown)


rhnonose:
We should be careful with Google Map APIs return, a strict validation thing will error on the user and reduce our opportunity to fix manually and document exceptions.

For now it's fine to keep things without strict validation since we can more easily fix manually with the dashboard.

pmargreff:
> For now it's fine to keep things without strict validation since we can more easily fix manually with the dashboard.

Ok, it looks reasonable to keep things like this. :+1:

apps/re/lib/addresses/address.ex
```diff
@@ -55,4 +56,22 @@ defmodule Re.Address do
   def generate_slugs(changeset) do
     Enum.reduce(@sluggified_attr, changeset, &Slugs.generate_slug(&1, &2))
   end
+
+
+  def truncate_postal_code(changeset) do
```
joaodubas: I think this operation isn't truncate, you're expanding the postal code to have 8 chars.

pmargreff: Renamed the function based in [`String.pad_trailing/3`](https://hexdocs.pm/elixir/String.html#pad_trailing/3). 

Is this more meaningful? If it isn't I'm open for suggestions.

joaodubas: Good one @pmargreff LGTM 👍

apps/re/test/addresses/address_test.exs
```diff
@@ -74,9 +74,11 @@ defmodule Re.AddressTest do
     assert changeset.errors == [postal_code: {"has already been taken", []}]
   end
 
-  test "five digits posta code should be valid" do
+  test "five digits postal code should be truncated to eight digits" do
```
joaodubas:
```suggestion
  test "five digits postal code should be padded to eight digits" do
```

pmargreff: Thank you, I missed the change here.

PR #407, opened 2019-02-04
Verify postal code with mix task

pmargreff:
This task complements PR #405.

The idea here is to iterate through the old listings and update the postal codes with 5 digits, when we found an existent postal code we point to the existent address, otherwise, we update the actual address. 

Once we shouldn't run every time (ideally only once) I think it's better to be a task model instead follow the migration model.

We tested this with a production based backup and also ran in staging and everything goes smoothly.

### DIFF
--- a/apps/re/lib/mix/tasks/fix_invalid_postal_codes.ex
+++ b/apps/re/lib/mix/tasks/fix_invalid_postal_codes.ex
@@ -1,9 +1,12 @@
 defmodule Mix.Tasks.Re.FixInvalidPostalCodes do
   use Mix.Task
   import Mix.Ecto
+  require Logger
 
   alias Re.{
     Address,
+    Addresses,
+    Listing,
     Listings.Queries,
     Repo
   }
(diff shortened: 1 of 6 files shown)


apps/re/test/mix/tasks/fix_invalid_postal_codes_test.exs
```diff
@@ -0,0 +1,61 @@
...
+      Mix.Tasks.Re.FixInvalidPostalCodes.run(nil)
+
+      listing = Repo.get(Listing, listing_id)
+
+      %{listings: [%{id: returned_listing_id}]} =
+        Repo.get(Address, valid_address_id)
+        |> Repo.preload(:listings)
+
+      assert [listing_id] == [returned_listing_id]
+      assert ^valid_address_id = listing.address_id
```
rhnonose: This should be a straight comparison rather than a match:

`assert valid_address_id == listing.address_id`.

When an assert fails, it doesn't use ExUnit's diff, so it's visually harder to interpret the error.

apps/re/test/mix/tasks/fix_invalid_postal_codes_test.exs
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    test "should delete old listing if is the only one" do
```
rhnonose: There's nothing that actually "deletes" anything, so this sounds like a misleading name for the test.

apps/re/lib/mix/tasks/fix_invalid_postal_codes.ex
```diff
@@ -0,0 +1,68 @@
+defmodule Mix.Tasks.Re.FixInvalidPostalCodes do
```
rhnonose: This task name is rather generic.

It should be more specific like `Mix.Tasks.Re.Addresses.PadPostalCodeRightWithZeroes` since we can't "describe" that action in the function name.

apps/re/test/mix/tasks/fix_invalid_postal_codes_test.exs
pmargreff: You're right, we change the test but forgot about the description ...

pmargreff:
Thank you nonose, changes requested added.

PR #410, opened 2019-02-08
Restart tag system

rhnonose:
This is a PoC for using CRQS with `commanded` and event sourcing with `eventstore` for the tag management system.
The previous PR #403 broke staging due to wrong variables, resulting in `too_many_connections` problem with heroku's postgres.
After evaluating if we should use the same database or not, for the sake of possibly migrating the rest to CQRS/ES, I created a separate umbrella app with it's own repo/database.

Since it's a big PR, I'll detail some choices
- Commanded and related versions are not the latest due to compatibility with ecto (will update when we remove blockers for ecto v3)
- `vex`, `exconstructor` and `ex_machina` from commanded's conduit example seemed like a good match for a cleaner code
- Projections with ecto will allow us to isolate a better query model to keep the same database for read store
- The models are oversimplified, we should revisit some requirements before deploying
- It's still not clear if projections or event handlers should be the way to update listings with tags


apps/re_tags/lib/tags/validators/unique_name.ex
```diff
@@ -0,0 +1,12 @@
...
+
+  alias ReTags.Tags
+
+  def validate(value, _context) do
+    case Tags.by_name(value) do
+      {:ok, _tag} -> {:error, "has already been taken"}
+      _ -> :ok
+    end
+  end
+end
```
joaodubas: When we update to Ecto 3.x, maybe we can use the  [`:on_conflict` clause][0] in `Ecto.Repo.insert` to avoid doing this call to the database.

[0]: https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert/2-upserts

apps/re_tags/lib/tags/projections/tag.ex
```diff
@@ -0,0 +1,12 @@
+defmodule ReTags.Tags.Projections.Tag do
+  use Ecto.Schema
+
+  @primary_key {:uuid, :binary_id, autogenerate: false}
+
+  schema "tag_projections" do
+    field :name, :string
```
joaodubas: If we want to avoid duplicate tag names, it makes sense to add a unique constraint in this field.

apps/re_tags/test/support/factory.ex
```diff
@@ -0,0 +1,17 @@
+defmodule ReTags.Factory do
+  use ExMachina
+
+  alias ReTags.Tags.Commands.CreateTag
+
+  def tag_factory do
+    %{
+      tag_uuid: UUID.uuid4(),
+      name: "naem",
```
joaodubas:
```suggestion
      name: "name",
```

apps/re_tags/lib/tags/validators/unique_name.ex
rhnonose: Since this is a command validation in an aggregate, we don't have access to `Repo` in the same manner as we normally would do.

Also, since event fields are not structured (probably serialized into jsonb or something), I'm pretty sure we can't use `on_conflict`.

apps/re_tags/lib/tags/projections/tag.ex
rhnonose: Since this is just a projection from created tags, the constraint here is not necessary.

apps/re_tags/test/support/factory.ex
rhnonose: It's wrong on purpose 😄

apps/re_tags/lib/tags/validators/unique_name.ex
joaodubas: I see, a shame we don't have access to this kind of construct, I think it makes the life easier.

I'm not sure if we can't define constraints in jsonb key/value pairs, will check the docs.

apps/re_tags/test/support/factory.ex
joaodubas: Nice 😄

apps/re_tags/lib/tags/validators/unique_name.ex
rhnonose: I think CQRS/ES wise you're not supposed to, because the aggregates are supposed to handle those constraints, but we'll discuss along the way.

rhnonose:
Closing to reconsider tag system with CQRS/ES

slashdotdash:
@rhnonose How did you find trying out Commanded? I'd be interested to hear your honest feedback, including any problems you encountered along the way. It's really useful feedback for me to see how I might be able to improve things, even if you've decided not to use it (for now).

rhnonose:
Hey @slashdotdash

One of the most visible is difficulty to do simple CRUDy stuff. Since we're still not knowledgeable in CQRS/ES, we were having some trouble designing our system to be event sourced, so naming was still hard and business rules were still nebulous and changing constantly.

The verbosity also was daunting, since one of the benefits of elixir is simple, short and elegant code, which as a little harder to write using `commanded`.

The thing is, we were still not sure if we needed a full-blown CQRS/ES, so we kinda decided to take some smaller steps with `ecto_job`, maybe do some event logging as well, which would suffice our needs.

But the most frustrating experience was setting up the test environment, which we still didn't manage to do completely (which you might've noticed since I spent a lot of time asking questions on slack's channel). Having both regular `Repo` and event-sourced stuff on test seemed almost impossible, so a gradual migration would be painful.

It's still not out of question to use in a next service we build, but we still feel it's quite a transition from the bread'n'butter, so it's a hard decision.

But thanks for reaching out for feedback.

slashdotdash:
Thanks for the feedback @rhnonose.

PR #411, opened 2019-02-12
Simple highlights heuristic

joaodubas:
This PR add logic to handle listings that are classified as highlights.

For the moment a highlight is a listing that:

1. Has up to 3 rooms.
2. Has at least 1 garage spots.
3. Has price up to R$ 2.000.000,00.
4. Located at a preferred neighborhood.

The highlight ordering is defined by two factors:

1. It's recency (the listing id corrected by the max listing id).
2. It's price per area (correct by the average price per area in the neighborhood).


apps/re/lib/listings/highlights_scores.ex
```diff
@@ -0,0 +1,76 @@
+defmodule Re.Listings.Highlights.Scores do
```
rhnonose: Usually, when the file is named `highlights_scores.ex`, the module name reflects as `HighlightsScores`.
I believe in this case it should be `highlights/scores.ex` and go inside it's own dir together with `highlights.ex`, otherwise it'll be cluttered under `listings` dir.

apps/re/lib/listings/highlights.ex
```diff
@@ -40,16 +41,31 @@ defmodule Re.Listings.Highlights do
...
+
+    all_highlights =
+      Queries.active()
+      |> Filtering.apply(filters)
+      |> Queries.preload_relations([:address])
+      |> Repo.all()
+      |> order_by_score()
+      |> Enum.drop(Map.get(params, :offset, 0))
+
+    if Map.has_key?(params, :page_size) do
```
rhnonose: You can extract this `if` block in a function and append to the pipeline for a cleaner code.
Instead of using a `if`, you can pattern match the presence of `page_size` key.

apps/re/lib/listings/highlights.ex
```diff
@@ -40,16 +41,31 @@ defmodule Re.Listings.Highlights do
...
-    filters = Map.get(params, :filters, %{})
-
-    Queries.active()
-    |> Filtering.apply(filters)
-    |> Queries.preload_relations([:address])
-    |> Queries.order_by(order)
-    |> Queries.limit(params)
-    |> Queries.offset(params)
-    |> Repo.all()
+    filters = mount_filters(params)
```
rhnonose: You can put this function logic together with `Filtering.apply` call inside `Scores` module since this is the only usage of the `Filtering` module.

apps/re/lib/listings/highlights.ex
```diff
@@ -40,16 +41,31 @@ defmodule Re.Listings.Highlights do
...
+      |> Enum.drop(Map.get(params, :offset, 0))
+
+    if Map.has_key?(params, :page_size) do
+      Enum.take(all_highlights, Map.get(params, :page_size, 0))
+    else
+      all_highlights
+    end
+  end
+
+  defp order_by_score(highlights) do
```
rhnonose: This function is not necessary, should be inlined as `|> Scores.calculate_highlights_scores()`

apps/re/test/listings/queries_test.exs
```diff
@@ -0,0 +1,63 @@
...
+
+      assert [] == result
+    end
+
+    test "average equals only listing for neighborhood" do
+      data = %{price: 10_000, area: 1_000, address: %{neighborhood_slug: "john-doe"}}
+
+      expected = [
+        %{
+          average_price_per_area: data.price / data.area,
```
rhnonose: This value should be hardcoded as well.

apps/re/test/listings/queries_test.exs
```diff
@@ -0,0 +1,63 @@
...
+        |> Repo.all()
+
+      assert expected == result
+    end
+
+    test "average for multiple neighborhoods" do
+      data_johndoe_1 = %{price: 10_000, area: 1_000, address: %{neighborhood_slug: "john-doe"}}
+      data_johndoe_2 = %{price: 15_000, area: 1_000, address: %{neighborhood_slug: "john-doe"}}
+      data_janedoe_1 = %{price: 20_000, area: 1_000, address: %{neighborhood_slug: "jane-doe"}}
+      data_janedoe_2 = %{price: 20_000, area: 2_000, address: %{neighborhood_slug: "jane-doe"}}
```
rhnonose: You should use a more appropriate name for a neighborhood (variable and data) rather than a person's name.

joaodubas: Updated the neighborhood names.

apps/re/lib/listings/highlights.ex
joaodubas: Removed the function and renamed the one in `Re.Listings.Highlights.Scores` to match it's purpose.

apps/re/lib/listings/highlights.ex
joaodubas: Extracted the query to it's own method and used pattern match to check if `page_size` was available.

rhnonose:
Also, run `mix credo --strict` and check the issues.

PR #413, opened 2019-02-13
Enable Vila Pompéia district

joaodubas:


apps/re/priv/repo/migrations/20190213173211_add_vila_pompeia_district.exs
```diff
@@ -0,0 +1,24 @@
+defmodule Re.Repo.Migrations.AddVilaPompeiaDistrict do
+  use Ecto.Migration
+
+  def up do
+    Re.Repo.insert(%Re.Addresses.District{
```
rhnonose: Here you'll have to use `District.changeset` to generate the slugs or put it manually.

PR #415, opened 2019-02-15
Add uuid for listing

rhnonose:
For the upcoming architectural changes in the project, we're gonna need entities to migrate to `uuid`.

We'll gradually create the fields and generate on insertion and migrate the relationships as needs emerge. For now, we'll spin off with listing for some external references.

### DIFF
--- a/apps/re/priv/repo/migrations/20190215015821_generate_listing_uuids.exs
+++ b/apps/re/priv/repo/migrations/20190215015821_generate_listing_uuids.exs
@@ -1,0 +1,26 @@
+defmodule Re.Repo.Migrations.GenerateListingUuids do
+  use Ecto.Migration
+
+  def up do
+    alter table(:listings) do
+      add :uuid, :string
+    end
+
+    flush()
+
+    Re.Listing
+    |> Re.Repo.all()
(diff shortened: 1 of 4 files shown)


apps/re/priv/repo/migrations/20190215015821_generate_listing_uuids.exs
```diff
@@ -0,0 +1,26 @@
...
+
+    flush()
+
+    create unique_index(:listings, [:uuid])
+  end
+
+  def down do
+    alter table(:listings) do
+      remove :uuid
+    end
```
joaodubas: The rollback should also remove the unique constraint.

rhnonose: For some reason I thought it removed together with the attribute, I'll double-check.

rhnonose: Seems like it, attempting to `drop index(:listings, [:uuid])` gives `** (Postgrex.Error) ERROR 42704 (undefined_object): index "listings_uuid_index" does not exist`

PR #418, opened 2019-02-19
Remove unecessary preload on related controller

pmargreff:
This PR aims to reduce timeouts from the production database, these looks to be caused by full preloads on listings (my suspects are on visualization assoc). 

I'll keep one eye on it and if it seems to reduce the timeouts will replicate this solution for all other get_preloaded use cases.

### DIFF
--- a/apps/re_web/lib/controllers/listings/related_controller.ex
+++ b/apps/re_web/lib/controllers/listings/related_controller.ex
@@ -7,10 +7,14 @@ defmodule ReWeb.RelatedController do
     Listings.Related
   }
 
+  @partial_preload [
+    :address
+  ]
+
   action_fallback(ReWeb.FallbackController)
 
   def index(conn, %{"listing_id" => id} = params, user) do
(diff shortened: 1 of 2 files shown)


apps/re_web/lib/controllers/listings/related_controller.ex
```diff
@@ -7,10 +7,14 @@ defmodule ReWeb.RelatedController do
...
 
+  @partial_preload [
+    :address
+  ]
+
   action_fallback(ReWeb.FallbackController)
 
   def index(conn, %{"listing_id" => id} = params, user) do
-    with {:ok, listing} <- Listings.get_preloaded(id),
+    with {:ok, listing} <- Listings.get_partial_preloaded(id, @partial_preload),
```
rhnonose: We have to serve the full preload for admin users.

rhnonose: You're right, it's not used in this context.

PR #419, opened 2019-02-21
Duplicate schema migrations table

rhnonose:
Since `eventstore` doesn't allow schema migration table customization, we need to rename our current `schema_migrations` table. This is a PR to discuss doing it through a migration or just manually running a SQL with a new configuration.


pmargreff:
Maybe it's a good idea to suggest this feature on `eventstore` repo as well.

rhnonose:
@pmargreff there's an issue there for some time but it's not a trivial PR since the migration scripts are `.sql` files.

rhnonose:
Closed in favor of #420

PR #420, opened 2019-02-21
Rename schema migration table

rhnonose:
This is an alternative to #419 
Before deploying this modification, we would take down the app and rename the table manually.
No migration fuss, plain and simple.


rhnonose:
Since heroku runs the `release` command on `Procfile` even with dynos scaled down to `0`, we had to disable for this PR and subsequent deploy.

Will be put back later.

PR #422, opened 2019-02-21
Export listings for all cities in trovit integration

joaodubas:
Maybe Trovit integration won't be segregated by state/city, so we export all listings in `/exporters/trovit`.


apps/re_web/lib/exporters/trovit/plug.ex
```diff
@@ -16,25 +16,37 @@ defmodule ReWeb.Exporters.Trovit.Plug do
...
-      |> Trovit.export_listings_xml()
-
-    conn
-    |> put_resp_content_type("application/xml")
-    |> send_resp(200, xml_listings)
+    response_with_xml_listings(filters, query_params, conn)
   end
 
-  def call(conn, _args) do
+  def call(%Plug.Conn{path_info: [_h|_t], query_params: _} = conn, _args) do
```
joaodubas: I'm not sure if this is the best approach to match any paths.

pmargreff:
```suggestion
  def call(conn, _args) do
```

Yes, give you one option, once we don't rely on these params we don't need to match them.

apps/re_web/lib/exporters/trovit/plug.ex
```diff
@@ -16,25 +16,37 @@ defmodule ReWeb.Exporters.Trovit.Plug do
...
       |> XmlBuilder.generate(format: :none)
 
+    response_for_exporter(error_response, conn, 404)
+  end
+
+  def call(%Plug.Conn{query_params: query_params} = conn, _args) do
+    response_with_xml_listings(%{}, query_params, conn)
+  end
+
+  defp response_with_xml_listings(filters, query_params, conn) do
```
pmargreff: Remove this function and duplicate the code in both call entries it's one option, I guess it has better readability and removes unnecessary abstraction (once `fetch_xml_listings` and `response_for_exporter` already isolate the implementations by themselves). 

However, I can recognize it can be an endless discussion, so feel free to ignore this comment if you disagree. :grimacing:

joaodubas: The initial implementation followed this suggestion, I thought it was better to reduce code duplication, and isolate each responsibility in it's own method. But I don't have strong opinions in favor of one or another.

As for the suggestion, I don't think it will work, since I must do the following:

1. Access to `/exporters/trovit/sp/sao-paulo/`: return _listings_ from São Paulo city
2. Access to `/exporters/trovit/`: return _listings_ from all cities
3. Access to `/exporters/trovit/anything/`: return 404 error
4. Access to `/exporters/trovit/sp/sao-paulo/lapa/`: return 404 error

joaodubas: > 
> 
> I'll suggest removing the call entry for the state/city match if they don't accept multiple URLs (in a future PR), so we don't keep any unused code.
> 
> Give some suggestions, in general, I guess it's good to go. 👍

You're absolutely right in here, if the target system won't filter by city, it doesn't make sense to keep the old the code (and [the best code ][0][is no code][1]).

[0]: https://blog.codinghorror.com/the-best-code-is-no-code-at-all/
[1]: https://github.com/kelseyhightower/nocode

joaodubas:
Trovit is able to handle multiple urls, so this PR is unnecessary.

PR #424, opened 2019-02-22
Add land development structure

pmargreff:
First version for land development for early feedback in some points I'm not 100% comfortable.   

- We mapped "empreendimentos" to development, which is a common name to this kind of venture but sounds too wide for me. Another common option is `land development` (this one sounds too raw for me), but it's another way to go, any thoughts on it? 

- Another point is what we call `status`. This one could be launch phase or even phase (status could be inactive/active related). Which one makes more sense for you guys?  
---
Also, I'll refactor images before merging this PR, once on our current state one image needs point to a listing, which will not be true anymore after this PR. 

Last but not least, I'll open the GraphQL queries in another PR, I guess I'll take some discovery time while doing it.

### DIFF
--- a/apps/re/lib/addresses/address.ex
+++ b/apps/re/lib/addresses/address.ex
@@ -24,6 +24,7 @@ defmodule Re.Address do
     field :state_slug, :string
 
     has_many :listings, Re.Listing
+    has_many :developments, Re.Development
 
     timestamps()
   end
--- a/apps/re/lib/developments/development.ex
+++ b/apps/re/lib/developments/development.ex
@@ -15,8 +15,8 @@ defmodule Re.Development do
     field :description, :string
 
     belongs_to :address, Re.Address
-
     has_many :images, Re.Image
+
     timestamps()
   end
 
--- a/apps/re/lib/developments/development.ex
+++ b/apps/re/lib/developments/development.ex
@@ -10,7 +10,7 @@ defmodule Re.Development do
   schema "developments" do
     field :name, :string
     field :title, :string
-    field :status, :string
+    field :phase, :string
     field :builder, :string
     field :description, :string
 
@@ -20,16 +20,16 @@ defmodule Re.Development do
     timestamps()
   end
 
-  @statuses ~w(pre-launch planning building delivered)
+  @phases ~w(pre-launch planning building delivered)
 
-  @required ~w(name title status builder description address_id)a
+  @required ~w(name title phase builder description address_id)a
 
   def changeset(struct, params) do
     struct
     |> cast(params, @required)
     |> validate_required(@required)
-    |> validate_inclusion(:status, @statuses,
-      message: "should be one of: [#{Enum.join(@statuses, " ")}]"
+    |> validate_inclusion(:phase, @phases,
+      message: "should be one of: [#{Enum.join(@phases, " ")}]"
     )
   end
 end
--- a/apps/re/lib/developments/development.ex
+++ b/apps/re/lib/developments/development.ex
@@ -1,0 +1,33 @@
+defmodule Re.Development do
+  @moduledoc """
+  Module for land developments.
+  """
+
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
+  schema "developments" do
+    field :name, :string
+    field :title, :string
+    field :status, :string
+    field :builder, :string
+    field :description, :string
+    belongs_to :address, Re.Address
+
+    timestamps()
+  end
+
+  @statuses ~w(pre-launch planning building delivered)
+
+  @required ~w(name title status builder description address_id)a
+
+  def changeset(struct, params) do
+    struct
+    |> cast(params, @required)
+    |> validate_required(@required)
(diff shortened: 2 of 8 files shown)


pmargreff:
I guess it's good enough to merge. 

Another point to discuss, the image <-> development assoc; At first I thought about something like a polymorphic association, [exploring the possible solutions](https://hexdocs.pm/ecto/2.1.4/Ecto.Schema.html#belongs_to/3-polymorphic-associations) I opted for the simple one, probably some time we will need to migrate for something more robust and flexible, but in meantime I think it's good enough.

pmargreff:
@rhnonose made the tests. Another point, should us validate at least one constraint (listing_id or development_id on this case) to create an image?

pmargreff:
The build does not agree with mine affirmation, taking a look at the error.

pmargreff:
@rhnonose build fixed, we're good to go again.

apps/re/lib/addresses/address.ex
```diff
@@ -24,6 +24,7 @@ defmodule Re.Address do
     field :state_slug, :string
 
     has_many :listings, Re.Listing
+    has_many :developments, Re.Development
```
joaodubas: This could also be `EstateDevelopment` or `EstateProject` (https://www.linguee.pt/portugues-ingles/traducao/empreendimento+imobiliário.html).

apps/re/lib/images/image.ex
```diff
@@ -13,6 +13,7 @@ defmodule Re.Image do
     field :description, :string
 
     belongs_to :listing, Re.Listing
+    belongs_to :development, Re.Development
```
joaodubas: In the future we could move both associations to different structs, something like:

```elixir
defmodule Re.ListingImage do
  belongs_to :image, Re.Image
  belongs_to: :listing, Re.Listing
end

defmodule Re.DevelopmentImage do
  belongs_to :image, Re.Image
  belongs_to :develoment, Re.Development
end
```

apps/re/lib/developments/development.ex
```diff
@@ -0,0 +1,35 @@
...
+  """
+
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
+  schema "developments" do
+    field :name, :string
+    field :title, :string
+    field :status, :string
```
joaodubas: Maybe `phase` or `construction_status` are better suited for this. Beside that, we could have a list of possible values for this field.

apps/re/test/support/factory.ex
```diff
@@ -143,6 +143,16 @@ defmodule Re.Factory do
 
   def tour_appointment_factory, do: %Re.Calendars.TourAppointment{}
 
+  def development_factory do
+    %{
+      name: Name.name(),
+      title: Name.name(),
+      status: Enum.random(~w(pre-launch planning building delivered)),
```
joaodubas: This list should be available in `Re.Development` struct. We should set a common pattern, [`Re.Listing` use both english and portuguese for this kind of data][0].

[0]: https://github.com/emcasa/backend/blob/4d0aa97ca33e059a6735aaee39022d4707a07dde/apps/re/lib/listings/schemas/listing.ex#L55-L57

apps/re/lib/images/image.ex
joaodubas: Forget about this, with `Event Source` this won't be needed.

joaodubas: How do we classify a given image, or this issue will be solved in another PR?

apps/re/lib/developments/development.ex
pmargreff: Thank you, I'll rename it.
About the status list, it's in [line 23 in this same file](https://github.com/emcasa/backend/pull/424/files#diff-09252c9ec11d6142b678f04888d933faR23).

apps/re/lib/images/image.ex
pmargreff: I choose to not solve this problem yet, we'll be able to categorize images only in the future.

apps/re/test/support/factory.ex
pmargreff: I pointed out the status list in another comment. About the common pattern I completely agree, use English would be the best way in my opinion. Types ones which use Portuguese definitions, it would be the easiest to change.

apps/re/lib/addresses/address.ex
pmargreff: Once we're already on this context Real Estate words would be too redundant (it's my guess). Most of our benchmarks use only development` (or `development and building`), which make some sense considering the implicit context. 

Make some parallel, our listings also could be `RealStateListing`, but once we're on this context it's not necessary, this argument makes sense?

joaodubas: Yeah, it makes sense, our product is related to real state, so we can imply that every concept belongs to real state.

apps/re/lib/images/image.ex
joaodubas: :ok:, this is a good approach, let's solve just one problem 👍

PR #427, opened 2019-02-25
Fix integration reported errors

joaodubas:
VivaReal and Zap provide an integration report, where any errors are detailed. This PR fixes errors presented in both providers.

Task: https://www.notion.so/emcasa/Corre-es-na-integra-o-Zap-VivaReal-5aeb6ac549b34ba3bca6b24419cda581

### DIFF
--- a/apps/re/lib/exporters/vivareal.ex
+++ b/apps/re/lib/exporters/vivareal.ex
@@ -178,6 +178,10 @@ defmodule Re.Exporters.Vivareal do
 
   defp translate_type("Apartamento"), do: "Apartment"
 
+  defp translate_type("Casa"), do: "Home"
+
+  defp translate_type("Cobertura"), do: "Penthouse"
+
   defp translate_type(_), do: "Other"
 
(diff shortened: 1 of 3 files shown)


joaodubas:
> I didn't saw the bathrooms problem addressed here, this PR should fix this problem as well @joaodubas?
> About the listing types, another option is to expose the listing types in a function and define the assertion dynamically, so we don't have surprises every time a new listing type is added and can find this early on tests. What are your thoughts about that?

1. The `bathrooms` and, maybe the `rooms`, should be fixed in the registration process. Remove these listings would be a mistake, since it just hides problems related to registration.
2. About the tests, expose the `Listing.@types` is the first step, the second is expose the _mapping_, maybe every exporter could expose this _mapping_. What do you think about that?

joaodubas:
For the bathrooms, I just warned everyone in _slack_, at our scale, this can be a possible solution. For the future, I really think, a solution would be validate our inputs during in **garagem**.

For the listing type, I will implement these methods and re-write the tests for all exporters :smile:!

PR #430, opened 2019-02-28
Mark non exportable listing

joaodubas:
There are sellers that won't allow us to list their realty in other sites. To control this behavior, we're adding a new column to `Re.Listing`, that will be used to determine if a given listing can be show in integration or not. This flag is editable by both admin and listing owner.

### DIFF
--- a/apps/re/priv/repo/migrations/20190215015821_generate_listing_uuids.exs
+++ b/apps/re/priv/repo/migrations/20190215015821_generate_listing_uuids.exs
@@ -1,6 +1,8 @@
 defmodule Re.Repo.Migrations.GenerateListingUuids do
   use Ecto.Migration
 
+  require Ecto.Query
+
   def up do
     alter table(:listings) do
       add :uuid, :string
@@ -8,14 +10,16 @@ defmodule Re.Repo.Migrations.GenerateListingUuids do
 
     flush()
 
-    Re.Listing
+    Ecto.Query.from(l in "listings", select: l.id)
     |> Re.Repo.all()
-    |> Enum.map(&Re.Listing.uuid_changeset(&1, %{uuid: UUID.uuid4()}))
-    |> Enum.each(&Re.Repo.update/1)
+    |> Enum.map(
+      &Ecto.Query.from(l in "listings", where: l.id == ^&1, update: [set: [uuid: ^UUID.uuid4()]])
+    )
+    |> Enum.map(&Re.Repo.update_all(&1, []))
 
     flush()
 
-    create unique_index(:listings, [:uuid])
+    create(unique_index(:listings, [:uuid]))
   end
 
   def down do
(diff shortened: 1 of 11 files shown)


apps/re/test/support/factory.ex
```diff
@@ -44,7 +44,8 @@ defmodule Re.Factory do
       matterport_code: Faker.String.base64(),
       status: "active",
       is_exclusive: Enum.random([true, false]),
-      is_release: Enum.random([true, false])
+      is_release: Enum.random([true, false]),
+      is_exportable: true
```
pmargreff: This value could be random as well. It allows us to see eventual interferences from this field in unrelated places...

apps/re/priv/repo/migrations/20190215015821_generate_listing_uuids.exs
```diff
@@ -1,21 +1,25 @@
...
+
   def up do
     alter table(:listings) do
       add :uuid, :string
     end
 
     flush()
 
-    Re.Listing
+    Ecto.Query.from(l in "listings", select: l.id)
```
pmargreff: nice catch :trophy:

apps/re/test/support/factory.ex
joaodubas: Nice, will change this one!

joaodubas: The only problem with this, is you must remember to set `is_exportable` in every integration test.

apps/re/priv/repo/migrations/20190215015821_generate_listing_uuids.exs
rhnonose: What's the issue here?

We shouldn't be altering migrations that's already in production unless there's an explicit problem like compiling or local migration.

joaodubas: Since I create a new migration, that add the `is_exportable` column, I'm not able to run the tests, or apply all migrations to a clean database, because of the following error (https://travis-ci.org/emcasa/backend/builds/500030391#L1550-L1558):

```bash
** (Postgrex.Error) ERROR 42703 (undefined_column): column l0.is_exportable does not exist
    (ecto) lib/ecto/adapters/sql.ex:431: Ecto.Adapters.SQL.execute_and_cache/7
    (ecto) lib/ecto/repo/queryable.ex:133: Ecto.Repo.Queryable.execute/5
    (ecto) lib/ecto/repo/queryable.ex:37: Ecto.Repo.Queryable.all/4
    apps/re/priv/repo/migrations/20190215015821_generate_listing_uuids.exs:12: Re.Repo.Migrations.GenerateListingUuids.up/0
    (stdlib) timer.erl:197: :timer.tc/3
    (ecto) lib/ecto/migration/runner.ex:25: Ecto.Migration.Runner.run/6
    (ecto) lib/ecto/migrator.ex:128: Ecto.Migrator.attempt/6
    (ecto) lib/ecto/migrator.ex:72: anonymous fn/4 in Ecto.Migrator.do_up/4
```

`Re.Listing` expects that the column is already available, but at this point, the database table doesn't have the column, and so, the data migration fails. That was the reason to change data migration to avoid any dependencies with the schema.

PR #431, opened 2019-02-28
Exposes development to graphql

pmargreff:
Exposing development query to Graphql, I used the plural to exposes the lists and the singular a single resource. This PR only exposes the queries, will submit the mutations on another PR.

Also made other minor changes, removed an unused function (`Re.Listings.index`), and renamed `ReWeb.Resolvers{Images, Address}.is_admin/3` to `ReWeb.Resolvers{Images, Addresses}.admin_rights?/3`. My expectation on rename is to give a better visibility about what it really checks and follow [Elixir name conventions](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/pages/Naming%20Conventions.md#trailing-question-mark-foo).

### DIFF
--- a/apps/re_web/lib/graphql/resolvers/addresses.ex
+++ b/apps/re_web/lib/graphql/resolvers/addresses.ex
@@ -9,13 +9,13 @@ defmodule ReWeb.Resolvers.Addresses do
   }
 
   def per_listing(listing, params, %{context: %{current_user: current_user}}) do
-    admin? = admin_rights?(listing, current_user)
+    admin? = admin_permissions?(listing, current_user)
 
     {:address, Map.put(params, :has_admin_rights, admin?)}
   end
 
   def per_development(_development, params, %{context: %{current_user: current_user}}) do
-    admin? = admin_rights?(nil, current_user)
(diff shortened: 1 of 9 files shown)


apps/re_web/lib/graphql/resolvers/addresses.ex
```diff
@@ -9,7 +9,13 @@ defmodule ReWeb.Resolvers.Addresses do
   }
 
   def per_listing(listing, params, %{context: %{current_user: current_user}}) do
-    admin? = is_admin(listing, current_user)
+    admin? = admin_rights?(listing, current_user)
```
joaodubas: Maybe `admin_permissions?` is a better name for this method.

joaodubas:
> How do we create/update a development?

Forget about this! You'll work on this in another PR.

apps/re_web/lib/graphql/resolvers/addresses.ex
rhnonose: Also maybe `has_admin_rights?` since it's the named passed down to the queries.

PR #432, opened 2019-03-01
Add graphql insert mutation to development

pmargreff:
This PR adds support to development changes through GraphQL. It adds both insert and update mutations, at beginning all operations will be handled only by admins as well visualization. 

Other related changes are:  
-It also creates other related things as Policy and the module to handle Development persistence, made it up based on the existing listing structure.
-Add some shared assertions for absinthe in a specific module.  

Sorry about the PR size, when I started, didn't suspect these changes would be so big.


apps/re_web/lib/graphql/types/development.ex
```diff
@@ -27,6 +27,17 @@ defmodule ReWeb.Types.Development do
...
   end
 
+  input_object :development_input do
+    field :name, :string
+    field :title, :string
+    field :phase, :string
+    field :builder, :string
+    field :description, :string
+
+    field :address, :address_input
```
rhnonose: I'm not sure about this `address_input` usage.
I reckon it's on `listing_input` as well, but now that I'm taking a closer look it might be confusing to have both as input and ID.
We should either document it properly (not sure where) or give an error if both are filled.
Ideally, we should remove the `input` portion and use only through ID.

apps/re_web/test/support/mutation_helpers.ex
```diff
@@ -216,4 +241,29 @@ defmodule ReWeb.Listing.MutationHelpers do
...
+          }
+          name
+          title
+          phase
+          builder
+          description
+        }
+      }
+    """
+  end
```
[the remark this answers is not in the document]
pmargreff: WOW, I misunderstand the use case for this helper, I'll move it to the right place.

apps/re_web/lib/graphql/types/development.ex
pmargreff: I can agree, while we're adding more relations, mutations could be really complex adding it through as nested attributes. 

I will remove it from these mutations and check if the front end uses the `address_input` on listings, if yes I'll deprecate if not I'll remove (whatever, I'll do it in another PR).

rhnonose: Awesome.

pmargreff:
@rhnonose made the required changes, could you review it again?

apps/re_web/lib/graphql/resolvers/developments.ex
```diff
@@ -37,7 +37,6 @@ defmodule ReWeb.Resolvers.Developments do
     end
   end
 
-  defp get_address(%{address: address_params}), do: Addresses.insert_or_update(address_params)
```
rhnonose: Now you shouldn't need this function and you can make  `address_id` as `non_null`

apps/re_web/lib/graphql/types/development.ex
```diff
@@ -15,6 +15,7 @@ defmodule ReWeb.Types.Development do
     field :phase, :string
     field :builder, :string
     field :description, :string
+    field :address_id, :id
```
rhnonose: This should be an `address` type so the client can query on top of it.

apps/re_web/lib/graphql/resolvers/developments.ex
pmargreff: already removed, this version is outdated.

apps/re_web/lib/graphql/types/development.ex
pmargreff: Didn't get it, `address` is exposed (two lines below), could you go deeper on this point?

rhnonose: Ah, my apologies.
I saw the id inclusion and didn't see the address exposed below.
This attribute then is unnecessary since you can get the id expanding the address attribute below.

pmargreff: Now I've removed all unnecessary entries ...

PR #435, opened 2019-03-07
Zerofill postal codes on address insertion

pmargreff:
This PR should solve the problem with price suggestion failing when should zero-fill address postal code. 

The problem was introduced [on this PR](https://github.com/emcasa/backend/pull/405/files), and raises the following problem: 

I was changing the postal code after check if the address exists, so it could try to duplicate an address which leads a constraint violation. 

E.g: When adding an address with invalid postal code (`99999`) and the postal code filled with zeroes already exists (`9999-000`), the `insert_or_update` try to insert a new record cause it searched with the invalid (`99999`) but it fails to find an address, so it tries to insert and the value is updated with the zero-filled postal code (`99999-000`) on changeset, it leads to an error cause it already exists. 

If the description is confusing, the [tests could be more conclusive](https://github.com/emcasa/backend/compare/zerofill-postal-codes-on-address-insertion?expand=1#diff-2b2467ec5eebaa9a0e748854471fd849R281).  

I removed the change from changeset and making the postal code fix on `insert_or_update` function.


rhnonose:
Good catch, indeed doing it on the changeset messes with the `find_or_insert` logic.
One nitpick is the logic that is duplicated between `pad_right_postal_code_wih_zeros.ex` and `addresses.ex`.

Although it's still 2 instances, maybe we should extract the logic to a common place or just outright delete the mix task since it was a one-time usage, what do you think?

joaodubas:
My 2 cents, we should delete the task, after all it was more of a data migration than a task.

pmargreff:
> Although it's still 2 instances, maybe we should extract the logic to a common place or just outright delete the mix task since it was a one-time usage, what do you think?

Agree @rhnonose and @joaodubas, how it should not be necessary anymore I removed the task.

PR #436, opened 2019-03-11
Setup eventstore

rhnonose:
- Add and configure `eventstore`
- This PR is to a feature branch because:
  - Heroku's 20 connection pool limitation might be a blocker for introducing a separate connection pool
  - `eventstore` requires a default 3 connections to function and additional connection pool
  - We have more reads hitting the database now and the left out 2 connections are not enough to supply that
  - We'll keep this feature branch until the blockers are cleared


rhnonose:
Yes, we can have another database.
For now it's more overhead to have a different database because it's harder to gradually migrate.
Say if we create an aggregate/event stream for one of the tables/objects, we might still need to perform some joins with other entities and cross-database joins are hard (also not sure if it's supported by ecto).
If we can't find a way to gradually migrate, then it's definitely the case that we'll have a different database from the start, but it seems like we can do gradual.

PR #437, opened 2019-03-12
Switch UUID poc

rhnonose:
Core:
- Create `uuid` reference in `tour_appointments`
- Set `tour_appointments`'s listing uuid
- Also fill `listing_uuid` when creating `tour_appointment`

Extra:
- Remove unused attribute in `trovit.ex`
- Add `subdirectories` to root `.formatter.exs` to remove the necessity to have all rules there


apps/re_web/lib/graphql/resolvers/calendars.ex
```diff
@@ -14,10 +14,13 @@ defmodule ReWeb.Resolvers.Calendars do
          do: {:ok, Calendars.generate_tour_options(Timex.now(), 5)}
   end
 
-  def schedule_tour(%{input: params}, %{context: %{current_user: current_user}}) do
+  def schedule_tour(%{input: %{listing_id: listing_id} = params}, %{
+        context: %{current_user: current_user}
+      }) do
     with :ok <- Bodyguard.permit(Calendars, :schedule_tour, current_user, %{}),
+         {:ok, listing} <- Listings.get(listing_id),
```
pmargreff: Is this an option to `get_by_uuid`, so we can follow the impacts, and catch problems early, or do you think it's unnecessary right now?

apps/re/priv/repo/migrations/20190311234958_switch_uuid.exs
```diff
@@ -0,0 +1,41 @@
...
+
+    flush()
+
+    alter table(:tour_appointments) do
+      add :listing_uuid, references(:listings, column: :uuid, type: :uuid)
+    end
+
+    flush()
+
+    Re.Calendars.TourAppointment
```
pmargreff: We already had some problems using modules references inside migrations, once they're changed or removed the migrations need to be changed...  

On this case, some changes can break the migration, like remove `listing_id` from Tour Appointment schema, this one could be solved using raw SQL instead.

rhnonose: I'll try some things here, good call.

apps/re_web/lib/graphql/resolvers/calendars.ex
rhnonose: This is just a modification to get the whole listing to fill the `uuid` field.
Since this functionality is sort of one-off thing (call mutation, send e-mail), there's no direct uuid call apart from this flow. If we switch to using `uuid` here it'll require a modification in the frontend that might not be small.

apps/re/lib/listings/schemas/listing.ex
```diff
@@ -7,7 +7,7 @@ defmodule Re.Listing do
   import Ecto.Changeset
 
   schema "listings" do
-    field :uuid, :string
+    field :uuid, Ecto.UUID
```
joaodubas: Nice :+1:

apps/re/priv/repo/migrations/20190312141527_add_listing_uuid_tour_appointments.exs
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    flush()
+
+    execute("""
+      UPDATE tour_appointments AS ta
+      SET listing_uuid = l.uuid
+      FROM listings AS l
+      WHERE ta.listing_id = l.id
+    """)
```
joaodubas: Conseguimos usar as primitivas do `Ecto`, ao invés de recorrer a uma _query_?

rhnonose: Maybe, I'll try to find a way.

rhnonose: So, `Repo.update_all` doesn't allow using the date returned by the query, so this operation will have to be a raw SQL.

I think even if it worked there, since it loads the struct, it might still give an error because the code won't have the attribute anymore.

PR #438, opened 2019-03-12
Add unit

pmargreff:
A unit represents the attributes that could have different values in the same listing. 
 
Next steps are:  
- Insert and change units. 
- Allow listing to receive a unit, required params in units will be not required on the listing changeset. We also could work on duplicating the unit inside listings on this step.

### DIFF
--- a/apps/re/priv/repo/migrations/20190311142702_create_units.exs
+++ b/apps/re/priv/repo/migrations/20190311142702_create_units.exs
@@ -1,8 +1,8 @@
-defmodule Re.Repo.Migrations.CreateRealties do
+defmodule Re.Repo.Migrations.CreateUnits do
   use Ecto.Migration
 
   def change do
-    create table(:realties) do
+    create table(:units) do
       add :uuid, :uuid
       add :complement, :string
       add :price, :integer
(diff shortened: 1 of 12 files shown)


apps/re/priv/repo/migrations/20190311142702_create_units.exs
```diff
@@ -0,0 +1,27 @@
...
+      add :bathrooms, :integer
+      add :restrooms, :integer
+      add :area, :integer
+      add :garage_spots, :integer
+      add :garage_type, :string
+      add :suites, :integer
+      add :dependencies, :integer
+      add :balconies, :integer
+
+      add :listing_id, references(:listings)
```
rhnonose: We can also start duplicating the references with `uuid` for easier migration.

pmargreff: I will do in the future, as I need [your migration](https://github.com/emcasa/backend/pull/437/files#diff-88ed5ec891f8f537b5af7a6b1fa96210R5) to create the reference [as you've done.](https://github.com/emcasa/backend/pull/437/files#diff-dff45307335ea142f992b935cf0149d7R17)

PR #440, opened 2019-03-13
Add units in Listing graphql query

pmargreff:
* Add Units graphql schema.
	* Batch load units per listing.
	* Add missing timestamp in units schema.
	* Add units in listing schema test.


apps/re/lib/units/units.ex
```diff
@@ -0,0 +1,9 @@
+defmodule Re.Units do
```
rhnonose: Will this be called outside of the development context?
Maybe this should be under `development`, like `Re.Developments.Units`.

apps/re/lib/units/units.ex
```diff
@@ -0,0 +1,9 @@
...
+  @moduledoc """
+  Context module for unit
```
joaodubas: What does a unit represent?

pmargreff: This entity should represent a realty/real state, for builders is common one listing to group more than one realty, but in normal cases, it will have only one unit.

apps/re/lib/units/units.ex
pmargreff: Not by now, will share if we choose to migrate for commom listings.

pmargreff:
> Is unit in production already? ...

Yes, it is, but I can replace it on the public interface, remove the common id and uses the only UUID for units to start to test this approach. Will make this change on this branch.

apps/re/lib/units/units.ex
joaodubas: Maybe this could go in the `moduledoc`.

pmargreff: It's, [but it's inside `Re.unit`](https://github.com/emcasa/backend/blob/master/apps/re/lib/units/unit.ex#L2), the schema's owner.

apps/re/lib/units/units.ex
rhnonose: Then, it should be an "internal module".
If it's top-level as it is, it might imply that it's made to be called directly.

apps/re/lib/units/units.ex
rhnonose: It should be on the context.

As we try to model the system in a DDD-style structure (context is sort of phoenix's way of bounded context), the context boundary is the window for external clients, so it should be in the context's `moduledoc`s.

But it's fine for now as we don't have an extensive code documentation since we're not too aligned with the domain yet, but just to keep in mind.

apps/re/lib/units/units.ex
pmargreff: Sorry, I got it wrong at first, it will be used outside of Development context.

pmargreff:
@rhnonose I updated it to use UUID, [also imported UUID as a scalar type](https://github.com/absinthe-graphql/absinthe/wiki/Scalar-Recipes#uuid-using-ectouuid), so we use it as a first-class citizen instead of a string.

rhnonose:
Overall looks good to me.

Can you add some basic structure of `development` and `unit` in the seeds? I always forget, but we should give more attention to it.

PR #442, opened 2019-03-14
Add category to images

pmargreff:
This add category to images end (it will be used only for development's images).

@joaodubas this could be a good opportunity to uses the tag on the feature. 

It also changes the assertions helper for absinthe assertions.


apps/re_web/test/graphql/images/mutation_test.exs
```diff
@@ -1,5 +1,8 @@
 defmodule ReWeb.GraphQL.Images.MutationTest do
-  use ReWeb.ConnCase
+  use ReWeb.{
+    AbsintheAssertions,
```
rhnonose: Nice one, didn't realize there was a helper for that.

PR #443, opened 2019-03-14
Use sentry to track errors

joaodubas:
Disable honeybadger and use sentry to track errors.


joaodubas:
> A few considerations:
> The `config.exs` modification mentions `production` and `staging` as included environments, how should this behave during development? Do we need to setup `SENTRY_DSN` locally?

We don't need to set `SENTRY_DSN` locally, since all exceptions will be ignored in local environment.

> Also, did you test if the dependency works if you only set in the root `mix.exs` file?

Didn't check that condition, will do some tests.

> Also, is it explicitly required to add `sentry` to `extra_applications`? As of elixir 1.4, it should load automatically.

Just followed the documentation for [elixir client][0], but I'll remove it and see if still works.

[0]: https://docs.sentry.io/clients/elixir/

joaodubas:
I'll revert the dependencies updates, and open a separate PR with these changes.

rhnonose:
So, in the documentation it asks to add to `applications` rather than `extra_applications`.
Since elixir 1.4, this is automatic.

If the sentry config doesn't affect local development, it should be in `config/prod.exs` rather than `config/config.exs`.
If there's a problem with that (for testing or compile error or something), you should add the minimum configuration to `config.exs` and have the production specific configuration in `config/prod.exs` (such as `SENTRY_DSN` config).

joaodubas:
> So, in the documentation it asks to add to `applications` rather than `extra_applications`.
> Since elixir 1.4, this is automatic.

I'm testing this change.

> If the sentry config doesn't affect local development, it should be in `config/prod.exs` rather than `config/config.exs`.
> If there's a problem with that (for testing or compile error or something), you should add the minimum configuration to `config.exs` and have the production specific configuration in `config/prod.exs` (such as `SENTRY_DSN` config).

I change it to `config/config.exs` so I could test on local environment, but I'll change it to `config/prod.exs` before merge.

joaodubas:
I can set the `sentry` dependency only in root `mix.exs`, but without the `:sentry` in the `extra_applications`, I got the following error:

```
2019-03-14T20:34:39.016058Z [error] #PID<0.626.0> running ReWeb.Endpoint (cowboy_protocol) terminated                                                                          
Server: 138.197.74.121:4000 (http)                                                                                                    
Request: GET /                                                          
** (exit) exited in: GenServer.call(Sentry.TaskSupervisor, {:start_task, [#PID<0.626.0>, :monitor, {:nonode@nohost, #PID<0.626.0>}, {:erlang, :apply, [#Function<3.32061305/0 in Sentry.Client.do_send_event/3>, []]}], :temporary, nil}, :infinity)
    ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
```

PR #448, opened 2019-03-18
Configure in memory db

rhnonose:
Configure an in-memory event store for test-only.


rhnonose:
Sorry, wrong merging target 😅

rhnonose:
Also edited the description to be more clear.

PR #449, opened 2019-03-19
Update ecto to v3

rhnonose:
- Update ecto to v3
- Fix test assertions that differs, doesn't seem to impact any clients
- Update and configure timber packages to fit ecto v3


apps/re/mix.exs
```diff
@@ -33,25 +33,26 @@ defmodule Re.Mixfile do
 
   defp deps do
     [
-      {:ecto, "~> 1.0 or ~> 2.0 or ~> 2.1"},
-      {:postgrex, ">= 0.0.0"},
+      {:ecto_sql, "~> 3.0"},
```
joaodubas: Do we need to set `ecto`?

mix.lock
```diff
@@ -7,30 +7,32 @@
...
+  "hackney": {:hex, :hackney, "1.15.1", "9f8f471c844b8ce395f7b6d8398139e26ddca9ebc171a8b91342ee15a19963f4", [:rebar3], [{:certifi, "2.5.1", [hex: :certifi, repo: "hexpm", optional: false]}, {:idna, "6.0.0", [hex: :idna, repo: "hexpm", optional: false]}, {:metrics, "1.0.1", [hex: :metrics, repo: "hexpm", optional: false]}, {:mimerl, "~>1.1", [hex: :mimerl, repo: "hexpm", optional: false]}, {:ssl_verify_fun, "1.1.4", [hex: :ssl_verify_fun, repo: "hexpm", optional: false]}], "hexpm"},
+  "honeybadger": {:hex, :honeybadger, "0.10.3", "1d1176bb64a82e8987e53f1f48670181ed6f44fde4674c7a0c95e188f974ae5f", [:mix], [{:hackney, "~> 1.1", [hex: :hackney, repo: "hexpm", optional: false]}, {:jason, "~> 1.0", [hex: :jason, repo: "hexpm", optional: false]}, {:phoenix, ">= 1.0.0 and < 2.0.0", [hex: :phoenix, repo: "hexpm", optional: true]}, {:plug, ">= 1.0.0 and < 2.0.0", [hex: :plug, repo: "hexpm", optional: true]}], "hexpm"},
```
joaodubas: `honeybadger` was replaced by `sentry`

rhnonose:
Good catch, missed that one. Maybe it was during merging.

apps/re/mix.exs
rhnonose: It's probably a good idea. `ecto_sql` depends on `ecto`, but I'll explicitly set it.

joaodubas: The [zen of python](https://www.python.org/dev/peps/pep-0020/#the-zen-of-python) tells this:

> Explicit is better than implicit.

PR #452, opened 2019-03-21
Send phone instead of email in notification

rhnonose:


apps/re_integrations/lib/notifications/emails/user.ex
```diff
@@ -214,7 +214,7 @@ defmodule ReIntegrations.Notifications.Emails.User do
                   #{unless request.is_covered, do: "Área fora de cobertura"}<br>
                   Preço sugerido: #{suggested_price || "Rua não coberta"}")
     |> text_body("Nome: #{request.name}
-                  Email: #{request.email}
+                  PHone: #{user.phone}
```
joaodubas:
```suggestion
                  Phone: #{user.phone}
```

rhnonose: ayy 😅

PR #453, opened 2019-03-22
Add GraphQL insert mutation to units

pmargreff:
Add GraphQL mutation to insert new units, as well all dependent logic (policy, context insertion, etc), also add the following structural changes: 
- Add relation between unit and development. 
- Add status attribute on units. 

Another thing that isn't a requirement but we agree is good enough until all filters are reimplemented on units is: development listings will use the lowest unit price by now.

### DIFF
--- a/apps/re/lib/units/policy.ex
+++ b/apps/re/lib/units/policy.ex
@@ -1,0 +1,14 @@
+defmodule Re.Units.Policy do
+  @moduledoc """
+  Policy module for user permission on units
+  """
+
+  alias Re.{
+    User
+  }
+
+  def authorize(_, %User{role: "admin"}, _), do: :ok
+  def authorize(_, nil, _), do: {:error, :unauthorized}
+
+  def authorize(_, _, _), do: {:error, :forbidden}
+end
--- a/apps/re/lib/units/policy.ex
+++ b/apps/re/lib/units/policy.ex
@@ -3,9 +3,7 @@ defmodule Re.Units.Policy do
   Policy module for user permission on units
   """
 
-  alias Re.{
-    User
-  }
+  alias Re.User
 
   def authorize(_, %User{role: "admin"}, _), do: :ok
   def authorize(_, nil, _), do: {:error, :unauthorized}
--- a/apps/re/lib/units/units.ex
+++ b/apps/re/lib/units/units.ex
@@ -8,6 +8,7 @@ defmodule Re.Units do
   alias Ecto.Changeset
 
   alias Re.{
+    Listings,
     Repo,
     Unit
   }
@@ -32,9 +33,7 @@ defmodule Re.Units do
          %{price: listing_price} = listing
        )
        when is_nil(listing_price) or unit_price < listing_price do
-    Changeset.change(listing, price: unit_price)
-    |> Repo.update()
-
+    Listings.update_price(listing, unit_price)
     new_unit
   end
 
--- a/apps/re/lib/units/units.ex
+++ b/apps/re/lib/units/units.ex
@@ -9,6 +9,7 @@ defmodule Re.Units do
 
   alias Re.{
     Listings,
+    PubSub,
     Repo,
     Unit
   }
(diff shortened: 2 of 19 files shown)


apps/re/lib/units/units.ex
```diff
@@ -3,8 +3,40 @@ defmodule Re.Units do
...
 
   def query(_query, _args), do: Re.Unit
+
+  def insert(params, development, listing) do
+    %Unit{}
+    |> Changeset.change(development_uuid: development.uuid)
+    |> Changeset.change(listing_id: listing.id)
+    |> Unit.changeset(params)
+    |> Repo.insert()
+    |> update_listing_price(listing)
```
rhnonose: Perhaps this should be done asynchronously through `PubSub` like in email and price/status history `server.ex`, for example.

apps/re/test/support/factory.ex
```diff
@@ -172,7 +172,8 @@ defmodule Re.Factory do
       garage_type: Enum.random(~w(contract condominium)),
       suites: Enum.random(0..10),
       dependencies: Enum.random(0..10),
-      balconies: Enum.random(0..10)
+      balconies: Enum.random(0..10),
+      status: Enum.random(~w(active inactive))
```
rhnonose: I'm not sure about this random, might have some inconsistency when testing.
Example: when the lowest price unit is inactive, the second lowest priced active unit should be used for the price, right?

apps/re_web/lib/graphql/types/unit.ex
```diff
@@ -20,5 +22,35 @@ defmodule ReWeb.Types.Unit do
...
+    field :dependencies, :integer
+    field :balconies, :integer
+    field :status, :string
+    field :development_uuid, non_null(:uuid)
+    field :listing_id, non_null(:id)
+  end
+
+  object :unit_mutations do
+    @desc "Insert unit"
+    field :insert_unit, type: :unit do
```
rhnonose: Since `unit` belongs to a `development`, `add_unit` would be a better name.

apps/re/lib/units/policy.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Re.Units.Policy do
+  @moduledoc """
+  Policy module for user permission on units
+  """
+
+  alias Re.{
+    User
+  }
```
joaodubas: Since we're only creating one alias, this could be only `alias Re.User`

apps/re/lib/units/units.ex
pmargreff: Makes sense, I will change it.

apps/re/test/support/factory.ex
pmargreff: Yeah, both your example and your concern looks right to me, I'll revert this change.

apps/re_web/lib/graphql/types/unit.ex
pmargreff: I've no strong opinion about this one, the only thing is to create a pattern and follow it to not have N different ways to name things. 

Today we already have {entity}_insert, insert_{entity}, if I'm adding add_{entity}. So with this task I'm also adding a card to refactor the image to `add_image`, on my understanding it should follow the same rule from the unit, is this affirmation right or the context is different?

pmargreff:
@rhnonose would you review it again? 

Two points I didn't think is the best way to go, `Propagator` is not a meaningful name, in the sense that it could be too generalistic. Another problem is the tests on the server module, I isolated the logic from Server itself in a way that if I can test the logic without touch the Server, so it isn't clear for me what should I pursue here. Should I duplicate the tests from `Propagator.update_listing/2`, test only again input and output interface is good enough, or we have other strategies here which I'm not s?  

Also, `PubSub` is isolating the dependency and unifying the interfaces in a really elegant way, congrats.  :trophy:

rhnonose:
@pmargreff LGTM
If you want to do an integrated test, take a look at `listings_test.exs:280`, there's an example of spinning up the server to test all together.
I'm still unsure about that approach (the rest of the tests are still isolated), we can talk about it.

pmargreff:
Awesome, I added one test on the happy way and found a pattern matching error. 

I think this strategy is good by now, but I agree we can discuss the scalability and how to not mess tests responsibilities.

pmargreff:
@rhnonose @joaodubas  Also, this merge introduces the following warning on master: 
```
warning: "else" clauses will never match because all patterns in "with" will always match
lib/listings/units/server.ex:28
```
For me looks like it's a false positive, [digging on elixir issues looks like it will be solved on OTP 22](https://github.com/elixir-lang/elixir/issues/6738), but would you double check [this match](https://github.com/emcasa/backend/blob/master/apps/re/lib/listings/units/server.ex#L26-L37) anyway?

PR #454, opened 2019-03-25
Create uuid field for user and generate it

rhnonose:
Create `uuid` field and generate for `users` to use with `tour_appointments` event-sourcing migration.

### DIFF
--- a/apps/re/lib/accounts/user.ex
+++ b/apps/re/lib/accounts/user.ex
@@ -61,11 +61,8 @@ defmodule Re.User do
     )
     |> validate_required(@account_kit_required)
     |> unique_constraint(:account_kit_id)
-    |> generate_uuid()
-    |> unique_constraint(:uuid, name: :uuid)
+    |> Re.ChangesetHelper.generate_uuid()
   end
-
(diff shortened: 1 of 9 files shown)


apps/re/lib/accounts/user.ex
```diff
@@ -82,4 +87,10 @@ defmodule Re.User do
       false -> add_error(changeset, :email, "has invalid format", validation: :format)
     end
   end
+
+  defp generate_uuid(%{data: %{uuid: nil}} = changeset) do
+    Ecto.Changeset.change(changeset, %{uuid: UUID.uuid4()})
+  end
+
+  defp generate_uuid(changeset), do: changeset
```
joaodubas: I think we should put the `generate_uuid/1` into it's own module.

rhnonose: I thought about it, but since it's a code that we'll throw away once we migrate to `commanded`, I thought about keeping it as it is.

But it might make sense to extract the whole `uuid_changeset`, that'll be easier to reuse once we introduce as a command, will work on it.

PR #455, opened 2019-03-26
Tagging system: schemas and queries

joaodubas:
The following changes were introduced:

1. Implement the [tag schema][0] with the [proper relations][1] and [queries][2].
2. Adjust listing to [reference their tags][3].
3. Add queries to [filter listing by tag slug and uuid][4].
4. Allow seed script to [remove data from database restored from backup][5].
5. Move [`slugs` to external module][6].

[0]: https://github.com/emcasa/backend/blob/77f19892d344a9aa6c37579b702f18533ad46b82/apps/re/lib/tags/tag.ex
[1]: https://github.com/emcasa/backend/blob/77f19892d344a9aa6c37579b702f18533ad46b82/apps/re/lib/tags/listing_tag.ex
[2]: https://github.com/emcasa/backend/blob/tag-system-take-01/apps/re/lib/tags/tags.ex
[3]: https://github.com/emcasa/backend/blob/77f19892d344a9aa6c37579b702f18533ad46b82/apps/re/lib/listings/schemas/listing.ex#L62-L65
[4]: https://github.com/emcasa/backend/blob/77f19892d344a9aa6c37579b702f18533ad46b82/apps/re/lib/filtering/filtering.ex#L211-L231
[5]: https://github.com/emcasa/backend/blob/77f19892d344a9aa6c37579b702f18533ad46b82/apps/re/priv/repo/seeds.exs
[6]: https://github.com/emcasa/backend/blob/77f19892d344a9aa6c37579b702f18533ad46b82/apps/re/lib/slugs/slugs.ex


apps/re/lib/tags/listing_tag.ex
```diff
@@ -0,0 +1,28 @@
...
+  Model that resolve relation between listing and tag.
+  """
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
+  @primary_key false
+
+  schema "listings_tags" do
+    belongs_to :listing, Re.Listing, primary_key: true
```
rhnonose: Maybe we should reference listing with `uuid` already, what do you think?
It's still not primary key but is being maintained.

apps/re/priv/repo/migrations/20190319192638_create_listings_tags.exs
```diff
@@ -0,0 +1,10 @@
+defmodule Re.Repo.Migrations.CreateListingsTags do
+  use Ecto.Migration
+
+  def change do
+    create table(:listings_tags, primary_key: false) do
+      add(:listing_id, references(:listings), primary_key: true)
+      add(:tag_uuid, references(:tags, column: :uuid, type: :uuid), primary_key: true)
```
rhnonose: Did the formatter put these parenthesis there or it was manual?
We should keep it without it for consistency. It's declared as an exception in `.formatter.exs`.

apps/re/lib/tags/tags.ex
```diff
@@ -0,0 +1,54 @@
...
+    |> Repo.insert()
+  end
+
+  def update(tag, params) do
+    tag
+    |> change(params)
+    |> Repo.update()
+  end
+
+  defp change(instance, params) do
```
rhnonose: Is this function necessary? If you inline `Tag.changeset/2`, it's gonna be pretty much the same.

apps/re/lib/tags/tag.ex
```diff
@@ -0,0 +1,55 @@
...
+  def changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, @required)
+    |> validate_required(@required)
+    |> unique_constraint(:name_slug)
+    |> validate_inclusion(:category, @categories,
+      message: "should be one of: [#{Enum.join(@categories, ", ")}]"
+    )
+    |> generate_slugs()
+    |> generate_uuid()
```
rhnonose: This was recently extracted to `Re.ChangesetHelper.generate_uuid/1`, you should use it.

apps/re/lib/tags/tags.ex
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def get(id) do
+    case Repo.get(Tag, id) do
+      nil -> {:error, :not_found}
+      tag -> {:ok, tag}
+    end
+  end
+
+  def list_by_ids(uuids) do
```
pmargreff: A function `list_by_ids` who receive an `uuid` can be confusing to a newcomer, or us on the future. Until we make the transitions (after this we can discuss again), I suggest we would make `uuid` uses explicitly.

apps/re/test/support/factory.ex
```diff
@@ -178,6 +178,20 @@ defmodule Re.Factory do
     }
   end
 
+  def tag_factory do
```
pmargreff: I guess this factory will fail if, in the same test we instantiate two instances with the same name (random make it possible) the due to the database constraint, one option is to remove `name` and `name_slug` from the factory and enforce the user to pass these params.

apps/re/lib/listings/listings.ex
```diff
@@ -197,4 +200,13 @@ defmodule Re.Listings do
     |> Queries.preload_relations(@full_preload)
     |> Repo.all()
   end
+
+  def upsert_tags(listing, tag_uuids, user) do
```
pmargreff: `user` param is not used here.

apps/re/lib/listings/schemas/listing.ex
```diff
@@ -111,6 +116,12 @@ defmodule Re.Listing do
     |> Re.ChangesetHelper.generate_uuid()
   end
 
+  def changeset_update_tags(struct, tags) do
+    struct
+    |> cast(%{}, @admin_attributes)
+    |> put_assoc(:tags, tags)
```
pmargreff: this should not be on common changeset for admins instead?

rhnonose: It's ok to have separate changesets for separate "operations".
But in this case, it should be only about `put_assoc` and remove `cast` to other attributes. Otherwise, it should be in the same changeset for sure.

apps/re/lib/tags/listing_tag.ex
joaodubas: It makes sense, I'm changing this one to reference the uuid field from listing 👍

apps/re/priv/repo/migrations/20190319192638_create_listings_tags.exs
joaodubas: I put it manually, will remove them.

apps/re/lib/tags/tags.ex
joaodubas: You're right, will inline it in both calls.

joaodubas:
@rhnonose I will figure out a way to reduce the performance impact of this join in the graphql query. Do we have this problem in the system? Just wanna check the code.

rhnonose:
@joaodubas there's `listing -> related` relationship that has some performance hiccups from time to time, but couldn't find a decent way of working around it.

It's not the same problem (joins), but it's similar in the sense that there's an extra hit in the database, might be worth looking at.

apps/re/lib/tags/tag.ex
```diff
@@ -26,11 +27,13 @@ defmodule Re.Tag do
...
 
-  @required ~w(name category)a
+  @required ~w(name category visibility)a
 
   @sluggified_attr [:name]
 
-  @categories ~w(infrastructure location realty)
+  @categories ~w(infrastructure location realty view)
+
+  @visibilities ~w(all admin)
```
rhnonose: I think `public`, `private` and/or `protected` are better names for visibility.

apps/re/lib/tags/tag.ex
```diff
@@ -17,6 +17,7 @@ defmodule Re.Tag do
     field :name, :string
     field :name_slug, :string
     field :category, :string
+    field :visibility, :string
```
rhnonose: Tip: `default: "private"` (if you follow my naming suggestion below) (also to align with the db's schema).

PR #457, opened 2019-03-27
Export all listing images on FacebookAds exporter

gabriel-laet:
* Do not export listings without images
* Export all listing images (max 20 according to FB docs) for both formats: `RealEstate` and `Product`

Depends on: #456

### DIFF
--- a/apps/re/lib/exporters/facebook_ads/real_estate.ex
+++ b/apps/re/lib/exporters/facebook_ads/real_estate.ex
@@ -15,11 +15,15 @@ defmodule Re.Exporters.FacebookAds.RealEstate do
     options = merge_default_options(options)
 
     listings
+    |> Enum.filter(&has_image?/1)
     |> Enum.map(&build_node(&1, options))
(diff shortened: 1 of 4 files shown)


apps/re/lib/exporters/facebook_ads/real_estate.ex
```diff
@@ -117,15 +121,7 @@ defmodule Re.Exporters.FacebookAds.RealEstate do
...
-    first_image = Enum.at(images, 0)
-
-    {
-      "image",
-      %{},
-      [
-        {"url", %{}, escape_cdata("#{@image_url}/#{first_image.filename}")}
-      ]
-    }
+    images |> Enum.map(&build_image_node(&1)) |> Enum.slice(0..19)
```
pmargreff: You can run mix format to brake lines here. 

You can also isolate the max size to a constant (`@max_images` or something like that), so we don't keep magic numbers, and you can use `Enum.take` to get the exact part you need ...

PR #458, opened 2019-03-27
Update elixir version

rhnonose:


.tool-versions-e
```diff
@@ -1,2 +1,2 @@
-elixir 1.7.4
+elixir 1.8.1
 erlang 21.0
```
pmargreff: Can you update Erlang version here also? It's probably my fault, I was the last one to update but didn't update this file once I didn't look at it.

rhnonose:
@pmargreff apparently it's a backup file created by `asdf`, it's safe to ignore.

PR #459, opened 2019-03-27
Fix with syntax warning

rhnonose:


rhnonose:
I think it's a warning that popped up after updating the elixir version, so it's fine.

PR #461, opened 2019-03-28
Allow to add development related listings

pmargreff:
I've added a new schema to allow listing with other requirements, the same for the changeset, at first I consider create o new module like `Listings.Development` to put changesets, but I thought it would have low visibility. 

Some business rules: 
- In the first moment, these listings will not be exportable, once we don't know the better format to adapter them to the consumers and don't want to follow it by now. 
- We do not consider farol anymore for sorting (I'll open other PR removing the dependency for this field). 

I need to make tests about how the system as a whole will react with and fix the main problems well as some refactor before merging. But by now I think we have code enough to discuss the approach and some options in case we have less turbulent ways.

### DIFF
--- a/apps/re/lib/listings/listings.ex
+++ b/apps/re/lib/listings/listings.ex
@@ -93,10 +93,12 @@ defmodule Re.Listings do
 
   def insert(params, address, user, development) do
     %Listing{}
-    |> Changeset.change(development_uuid: development.uuid)
-    |> Changeset.change(address_id: address.id)
-    |> Changeset.change(user_id: user.id)
-    |> Changeset.change(is_exportable: false)
+    |> Changeset.change(%{
+      development_uuid: development.uuid,
+      address_id: address.id,
+      user_id: user.id,
+      is_exportable: false
+     })
     |> Listing.development_changeset(params)
     |> Repo.insert()
     |> publish_if_admin(user.role)
--- a/apps/re/lib/listings/listings.ex
+++ b/apps/re/lib/listings/listings.ex
@@ -84,9 +84,21 @@ defmodule Re.Listings do
   def get_partial_preloaded(id, preload),
     do: do_get(Queries.preload_relations(Listing, preload), id)
 
-  def insert(params, address, user) do
+  def insert(params, address, user, development \\ nil)
+
+  def insert(params, address, user, nil) do
     with {:ok, user} <- validate_phone_number(params, user),
          do: do_insert(params, address, user)
+  end
+
+  def insert(params, address, user, development) do
+    %Listing{}
+    |> Changeset.change(development_uuid: development.uuid)
+    |> Changeset.change(address_id: address.id)
+    |> Changeset.change(user_id: user.id)
+    |> Changeset.change(is_exportable: false)
+    |> Listing.development_changeset(params)
+    |> Repo.insert()
   end
 
   defp do_insert(params, address, user) do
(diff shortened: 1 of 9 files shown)


apps/re/lib/listings/schemas/listing.ex
```diff
@@ -108,7 +108,22 @@ defmodule Re.Listing do
...
+
+  @development_attributes @development_required ++ @development_optional
+
+  def development_changeset(struct, params) do
+    struct
+    |> cast(params, @development_attributes)
+    |> cast_assoc(:development)
+    |> validate_required(@development_required)
+    |> validate_inclusion(:type, @types, message: "should be one of: [#{Enum.join(@types, " ")}]")
+    |> generate_uuid()
```
rhnonose: Why not calling directly `Re.ChangesetHelper.generate_uuid/1`?

apps/re/lib/listings/listings.ex
```diff
@@ -84,11 +84,24 @@ defmodule Re.Listings do
...
     with {:ok, user} <- validate_phone_number(params, user),
          do: do_insert(params, address, user)
   end
 
+  def insert(params, address, user, development) do
+    %Listing{}
+    |> Changeset.change(development_uuid: development.uuid)
+    |> Changeset.change(address_id: address.id)
+    |> Changeset.change(user_id: user.id)
+    |> Changeset.change(is_exportable: false)
```
rhnonose: Maybe it's better to group those in a single `Changeset.change/2` call.

apps/re/lib/listings/schemas/listing.ex
pmargreff: Isolating dependencies, once we need to change public interfaces for `ChangesetHelper` we have only one place to change in this module.

rhnonose: But it's reducing 2 to 1, similar in other modules. Not sure if it's worth an extra function.

pmargreff: 3 actually, I have an opinion slightly different. While isolating dependencies, we should not see extra functions as extra cost/effort, I agree we shouldn't require all dependencies be encapsulated, but once they are I can't see downsides.

Anyway,  if it,s a window for the problem in the future, I can revert this change. Credo will be helpful to drive and enforce these discussions to find what are or not patterns to follow ...

pmargreff:
I guess it's good to review now. I change the approach to use the same GraphQL schemas we already use on common listings. 

Also added some development listings to see how the frontend reacts (banduntss). It was pretty good, only found a possible problem while search listings, already reported this to fix, and while we haven't any active development listing before this fix, everything should be good.

apps/re/lib/listings/policy.ex
```diff
@@ -10,6 +10,7 @@ defmodule Re.Listings.Policy do
 
   def authorize(_, %User{role: "admin"}, _), do: :ok
   def authorize(:create_listing, %User{role: "user"}, _), do: :ok
+  def authorize(:create_development_listing, %User{role: "admin"}, _), do: :ok
```
joaodubas: I don't think this match will be triggered!

pmargreff: I'm using it [here](https://github.com/emcasa/backend/pull/461/files#diff-b4d1e33dc452809ab3e2d9b0fbafbd17R48) and [testing it here](https://github.com/emcasa/backend/pull/461/files#diff-a0943423f2c83c7f36922322b01eae80R406). 

But if your warning was already considering these appointments, ping me, please.

PR #462, opened 2019-03-28
Fix unit nonconformities

pmargreff:
@gabiseabra pointed out these differences between unit and listings behaviors. 

- Both `garage_type` and `dependencies` should not be required. 
- Use enum on `garage_type` instead of plain strings. 

Fixing them with this PR.

### DIFF
--- a/apps/re_web/test/graphql/units/mutation_test.exs
+++ b/apps/re_web/test/graphql/units/mutation_test.exs
@@ -18,7 +18,8 @@ defmodule ReWeb.GraphQL.Units.MutationTest do
     unit_params =
       string_params_for(:unit, %{
         development_uuid: development.uuid,
-        listing_id: listing.id
+        listing_id: listing.id,
+        garage_type: "CONDOMINIUM"
       })
       |> Map.delete("uuid")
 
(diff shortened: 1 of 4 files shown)


apps/re_web/test/graphql/units/mutation_test.exs
```diff
@@ -79,7 +80,7 @@ defmodule ReWeb.GraphQL.Units.MutationTest do
       assert insert_unit["restrooms"] == unit_params["restrooms"]
       assert insert_unit["area"] == unit_params["area"]
       assert insert_unit["garage_spots"] == unit_params["garage_spots"]
-      assert insert_unit["garage_type"] == unit_params["garage_type"]
+      assert insert_unit["garage_type"] == String.downcase(unit_params["garage_type"])
```
rhnonose: Maybe it's worth migrating current filled fields to uppercase so we don't have to keep doing this conversion. What do you think?

pmargreff: Reflect the changes from our Public API to our internal structure sounds like a smell. But I really like the idea of enums where they are applicable, and this problem will be more evident through time. 

Another approach is to create an `insensitive_case_assert` and at the same time [extend factories](https://hexdocs.pm/ex_machina/readme.html#flexible-factories-with-pipes) with a function as `convert_enum_params` (one for each structure who exposes an enum, `convert_{listing, unit}_enum_parmas... ) so we isolate the enums creation. I've used this solution in the past and it has been useful.

PR #463, opened 2019-03-29
Improve CI

rhnonose:
- [x] Replace TravisCI with CircleCI
- [x] Fix coveralls
- [x] Add warnings as errors
- [x] Add formatter check
- [x] Add credo check


circle.yml
```diff
@@ -0,0 +1,56 @@
...
+
+      - restore_cache:
+          keys:
+            - v1-mix-cache-{{ .Branch }}-{{ checksum "mix.lock" }}
+            - v1-mix-cache-{{ .Branch }}
+            - v1-mix-cache
+      - restore_cache:
+          keys:
+            - v1-build-cache-{{ .Branch }}
+            - v1-build-cache
```
joaodubas: Why do we have two `restore_cache`  items?

circle.yml
```diff
@@ -0,0 +1,56 @@
...
+      - save_cache:
+          key: v1-build-cache-{{ .Branch }}
+          paths: "_build"
+      - save_cache:
+          key: v1-build-cache
+          paths: "_build"
+
+      - run:
+          name: Wait for DB
+          command: dockerize -wait tcp://localhost:5432 -timeout 1m
```
joaodubas: You could also use the [`pg_isready`][0] command to check connectivity.

[0]: https://www.postgresql.org/docs/10/app-pg-isready.html

circle.yml
```diff
@@ -0,0 +1,56 @@
...
+
+      - run: mix test
```
joaodubas: We could check the code coverage.

circle.yml
rhnonose: I got a default template 😂 
What do you recommend?

joaodubas: Just thought it was strange, but I can see the reason, we're separating dependency cache from build cache.

.circleci/config.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - run:
+          name: Wait for DB
+          command: dockerize -wait tcp://localhost:5432 -timeout 1m
+
+      - run: mix coveralls.circle --umbrella
+      - store_artifacts:
+          path: cover/excoveralls.html
+      - store_test_results:
+          path: _build/test/junit
```
joaodubas: Nice 👍

circle.yml
```diff
@@ -0,0 +1,63 @@
...
+      - store_test_results:
+          path: /tmp/backend-repor-test-result
+
+      - run: mix coveralls.circle --umbrella
+      - store_artifacts:
+          path: cover/excoveralls.html
+
+      - run: mix credo --strict
```
joaodubas: Is this file needed at all?

.circleci/config.yml
rhnonose: Not working 😢

circle.yml
rhnonose: I'll remove after adding stickler

.circleci/config.yml
rhnonose: It is now 😂

circle.yml
rhnonose: Seems unavailable.
Do you think it's worth installing or is it fine this way?

joaodubas: It's fine this way!

joaodubas: Probably this is the recommendation from circleci.

pmargreff:
Great job :+1: . 

Late feedbacks, coveralls badge doesn't seem to be updating (probably broked link?).

rhnonose:
@pmargreff coveralls was messed up for some reason, looks ok on master

PR #464, opened 2019-03-29
Add listings for development

pmargreff:
Allow the user to get all listings from a development. 

Also, changed the git hook, actually we're not formatting test files, but our `pre-commit` hook is taking it in accounts, so I'm excluding tests from on our `pre-commit` model, you'll need to create again or update in your local machine to bypass this problem. Is this reasonable for you?

### DIFF
--- a/apps/re_web/lib/graphql/types/development.ex
+++ b/apps/re_web/lib/graphql/types/development.ex
@@ -25,6 +25,8 @@ defmodule ReWeb.Types.Development do
 
       resolve &Resolvers.Images.per_development/3
     end
+
+    field :listings, list_of(:listing), resolve: &Resolvers.Listings.per_development/3
   end
 
   input_object :development_input do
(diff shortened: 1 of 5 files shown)


rhnonose:
@pmargreff I think it's a mistake that the test files are not included. I think I changed it when I was splitting the formatter configurations, it should be added.
`inputs: ["mix.exs", "lib/**/*.{ex,exs}"]` -> `inputs: ["mix.exs", "{lib,test}/**/*.{ex,exs}"]`
Also migrations as well, maybe we can do in another PR.

apps/re_web/lib/graphql/types/development.ex
```diff
@@ -25,6 +25,8 @@ defmodule ReWeb.Types.Development do
 
       resolve &Resolvers.Images.per_development/3
     end
+
+    field :listings, list_of(:listing), resolve: &Resolvers.Listings.per_development/3
```
rhnonose: I think you can use the "short-form" for loading this, like:
`field :listings, list_of(:listing), resolve: dataloader(Re.Listings)`
If it works, you won't need `per_development/3` function.
(might require adding `development` relationship in listing's schema if there isn't)

pmargreff: Awesome, the first try really looks like a lot of boilerplate for one common functionality. 

Thank you. :purple_heart:

PR #467, opened 2019-04-02
Allow edit development listings

pmargreff:
* Add GraphQL queries to edit development listings with
	right validations when updateListing mutation receives
	development_uuid params.
	* Remove create policy once the match will occur always at first
	 policy function.
	* Extract development listing input.
--- 

Also, add a new param for each association for both insert and update functions doesn't seem to be a good pattern, it tends to increase the params number as well to lead to not so flexible public APIs. My next step is to open a new PR refactoring those cases to use a map as the third param with associations and matching it against the right insert/update function, i.e., `def update(listing, params, %{user: user, address: address} = associations)`

Have you any other suggestions/options for this topic?

### DIFF
--- a/apps/re_web/lib/graphql/resolvers/listings.ex
+++ b/apps/re_web/lib/graphql/resolvers/listings.ex
@@ -74,6 +74,19 @@ defmodule ReWeb.Resolvers.Listings do
   defp get_development(%{development_uuid: uuid}), do: Developments.get(uuid)
   defp get_development(_), do: {:error, :bad_request}
 
+  def update(%{id: id, input: %{development_uuid: _} = listing_params}, %{
+        context: %{current_user: current_user}
+      }) do
+    with {:ok, listing} <- Listings.get(id),
+         :ok <- Bodyguard.permit(Listings, :update_development_listing, current_user, listing),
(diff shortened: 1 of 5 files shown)


apps/re_web/lib/graphql/resolvers/listings.ex
```diff
@@ -74,6 +74,19 @@ defmodule ReWeb.Resolvers.Listings do
   defp get_development(%{development_uuid: uuid}), do: Developments.get(uuid)
   defp get_development(_), do: {:error, :bad_request}
 
+  def update(%{id: id, input: %{development_uuid: _} = listing_params}, %{
+        context: %{current_user: current_user}
+      }) do
+    with {:ok, listing} <- Listings.get(id),
+         :ok <- Bodyguard.permit(Listings, :update_development_listing, current_user, listing),
+         {:ok, address} <- get_address(listing_params),
+         {:ok, development} <- get_development(listing_params),
```
rhnonose: There's 3 `get`s here, we should query once with a `preload`.

pmargreff: Ok, this will override the associations passed in params, but how they should be not editable directly by the client I see no problem.

PR #468, opened 2019-04-02
Filter listing by tag slug name or uuid on graphql query

joaodubas:
Add the option to search listings by list of tag slugs name or uuids. Example:

```graphql
query {
  listings(filters: {tagsSlug: ["salao-de-festas"]}) {
    listings {
      id
      address {
        street
        streetNumber
      }
    }
  }
}
```

### DIFF
--- a/apps/re/test/listings/queries_test.exs
+++ b/apps/re/test/listings/queries_test.exs
@@ -64,7 +64,7 @@ defmodule Re.Listings.QueriesTest do
   end
 
   describe "remaining_count/2" do
-    test "get remaining count with single tag search" do
+    test "remaining count with single tag search should bring correct number of listings" do
       tag_1 = insert(:tag, name: "Tag 1", name_slug: "tag-1")
       insert(:listing, tags: [tag_1])
       insert(:listing, tags: [tag_1])
@@ -78,7 +78,7 @@ defmodule Re.Listings.QueriesTest do
       assert 2 == result
     end
 
-    test "get remaining count with multi tag search" do
+    test "remaining count with multi tag search should bring correct number of listings, despite many to many relation" do
       tag_1 = insert(:tag, name: "Tag 1", name_slug: "tag-1")
       tag_2 = insert(:tag, name: "Tag 2", name_slug: "tag-2")
       insert(:listing, tags: [tag_1, tag_2])
--- a/apps/re/test/listings/queries_test.exs
+++ b/apps/re/test/listings/queries_test.exs
@@ -2,6 +2,8 @@ defmodule Re.Listings.QueriesTest do
   use Re.ModelCase
 
   alias Re.{
+    Filtering,
+    Listing,
     Listings.Queries,
     Repo
   }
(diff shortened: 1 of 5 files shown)


apps/re_web/test/graphql/listings/query_test.exs
```diff
@@ -305,6 +305,9 @@ defmodule ReWeb.GraphQL.Listings.QueryTest do
     end
 
     test "should query listing index with filtering", %{user_conn: conn} do
+      tag_1 = insert(:tag, name: "Tag 1", name_slug: "tag-1")
+      tag_2 = insert(:tag, name: "Tag 2", name_slug: "tag-2")
```
rhnonose: I think there's a way of inserting the relationship without calling `Listings.upsert_tags/2`on every listing, maybe you should check `ex_machina`'s documentation.
If `tags: [build(:tag), build(:tag)]` directly doesn't work, you can create a `tag_listing_factory`. It should shrink the test a little bit.

apps/re_web/lib/graphql/types/listing.ex
```diff
@@ -203,6 +203,8 @@ defmodule ReWeb.Types.Listing do
     field :garage_types, list_of(:garage_type)
     field :cities, list_of(:string)
     field :cities_slug, list_of(:string)
+    field :tags_slug, list_of(:string)
+    field :tags_uuid, list_of(:uuid)
```
rhnonose: These should be `list_of(non_null(:string))` and `list_of(non_null(:uuid))`.
The field can be `nil` but the elements shouldn't.

apps/re/test/listings/queries_test.exs
```diff
@@ -60,4 +60,9 @@ defmodule Re.Listings.QueriesTest do
       assert expected == result
     end
   end
+
+  describe "remaining_count/2" do
+    test "get remaining count with tags" do
+    end
+  end
```
rhnonose: Did forget this hanging?

joaodubas: Ops, my bad! End up testing this through `Re.Listings.paginated`.

apps/re_web/lib/graphql/types/listing.ex
joaodubas: Should I change the others attributes that are set to `list_of`?

rhnonose: If it's new and the clients didn't implement yet, yes.
Otherwise, it might break some queries that have specified the types in details.

joaodubas: 🆗

apps/re_web/test/graphql/listings/query_test.exs
joaodubas: You were right, rewrote some of the tests to add tags directly into listing, using the `insert` method.

apps/re/test/listings/queries_test.exs
```diff
@@ -60,4 +62,35 @@ defmodule Re.Listings.QueriesTest do
       assert expected == result
     end
   end
+
+  describe "remaining_count/2" do
```
pmargreff: These tests look touching too much. If they should test only `remaining_count` they don't need to touch attributes or other functions as filters. A simple case is to create some listings and test the remaining count. So we don't touch `Filtering.apply` neither tags. 

Otherwise, if the intention was to create these integration test, maybe the describe message isn't appropriated. Does this comment make sense?

joaodubas: They're a testing an edge case for the `Re.Listings.Queries.remaining_count/2`.

I'll improve the description for both tests.

joaodubas: This is an example of what this test ensure: http://sqlfiddle.com/#!17/50fec/3

PR #470, opened 2019-04-04
Add filter for status on listings

pmargreff:
Add status filter as requested by @gabiseabra. 

Using an array to allow multiple statuses, as our next steps will probably add more options in this field, and using arrays, we wouldn't break any public API... 

Also testing corner cases on filter module test and test integration (GraphQL request) in a different test block.


apps/re_web/lib/graphql/types/listing.ex
```diff
@@ -203,6 +203,7 @@ defmodule ReWeb.Types.Listing do
     field :garage_types, list_of(:garage_type)
     field :cities, list_of(:string)
     field :cities_slug, list_of(:string)
+    field :statuses, list_of(:string)
```
rhnonose: This should be `list_of(non_null(:string))`

pmargreff:
testing coveralls

apps/re_web/lib/graphql/types/listing.ex
pmargreff: So, shouldn't be other filters as well? Should we update them?

rhnonose: The trouble with updating other fields is that it might break the queries that defines the types as list of "might be null", so might require syncing with the clients.

PR #472, opened 2019-04-04
Cqrs es migrations

rhnonose:
This is the setup and simples PoC of our most common workflow in the system (CRUD -> PubSub -> Integration calls).

Benefits:
- Everything is an event, so we have a good source to analyse data
- We can tweak strong/eventual consistency as well as sync/async
- Consistency check in aggregates makes it easier to split the system
- You can more easily derive the system's behavior from the code
Drawbacks:
- `commanded` doesn't feel right (test problems, not handling disconnections well, boilerplate)
- The simplest feature feels bloated with too much code
- Team knowledge on CQRS/DDD/ES is still shallow
- Heroku's connection pool limitation requires us to migrate to somewhere else first (or use an expensive tier)
- Things are still changing, so event versioning might be a problem
- There's no dedicated mature storage for this model


rhnonose:
Closing this PR in favor of searching for an intermediate alternative.

Options:
- `ecto_job` for background processing with consistency
- Study DDD: learn what's important to be applied and don't go with the hype
- Event logging, domain events, ES without CQRS

PR #474, opened 2019-04-08
Grahpql insert images on development

pmargreff:
Allow insertion of different image types. It required some changes and I took the opportunity to add a more flexible interface for images, main changes are: 
- Now we have a parent_type, to allow be associated with multiple types, we can use pattern match to route the insertion to right resolver function. 
- Using uuid to identify the parent. 
- Now the parent is returned, added a new output param called `parent` by now it can be both `Development` or `Listing` union type, in the future, we can add more image associations (tags, user, etc..) without brake public API. 
- Removed `listing_id` as non-null once development images will not know this param. 

In the sequence I'll: 
- Make the same for image updates, activation and deactivation. 
- Add this same interface to handle listing images and deprecated the old one, so in a nearby future, we can turn both `parentType` and `uuid` non_null. 
- Will change the way we work with image subscription, notifying the association update, so instead of emitting an `images_update:{id}`, we'll emit `development_updated:uuid`, etc. 

Suggestions for other approaches are always welcome.

### DIFF
--- a/apps/re/lib/developments/queries.ex
+++ b/apps/re/lib/developments/queries.ex
@@ -1,0 +1,13 @@
+defmodule Re.Developments.Queries do
+  @moduledoc """
+  Module for grouping develpments queries
+  """
+
+  alias Re.Development
+
+  import Ecto.Query
+
+  def preload_relations(query \\ Development, relations \\ [])
+
+  def preload_relations(query, relations), do: preload(query, ^relations)
+end
--- a/apps/re/lib/developments/queries.ex
+++ b/apps/re/lib/developments/queries.ex
@@ -1,6 +1,6 @@
 defmodule Re.Developments.Queries do
   @moduledoc """
-  Module for grouping develpments queries
+  Module for grouping developments queries
   """
 
   alias Re.Development
--- a/apps/re_web/lib/graphql/resolvers/images.ex
+++ b/apps/re_web/lib/graphql/resolvers/images.ex
@@ -63,7 +63,7 @@ defmodule ReWeb.Resolvers.Images do
     with {:ok, development} <- Developments.get_preloaded(parent_uuid, [:images]),
          :ok <- Bodyguard.permit(Images, :create_development_images, current_user, development),
          {:ok, image} <- Images.insert(params, development, :development) do
-      {:ok, %{parent_listing: nil, image: image}}
+      {:ok, %{parent: development, image: image}}
     end
   end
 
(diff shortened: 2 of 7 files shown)


apps/re/lib/developments/queries.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Re.Developments.Queries do
+  @moduledoc """
+  Module for grouping develpments queries
```
joaodubas:
```suggestion
  Module for grouping developments queries
```

apps/re_web/lib/graphql/resolvers/images.ex
```diff
@@ -56,6 +57,16 @@ defmodule ReWeb.Resolvers.Images do
     end)
   end
 
+  def insert_image(%{input: %{parent_type: :development, parent_uuid: parent_uuid} = params}, %{
+        context: %{current_user: current_user}
+      }) do
+    with {:ok, development} <- Developments.get_preloaded(parent_uuid, [:images]),
+         :ok <- Bodyguard.permit(Images, :create_development_images, current_user, development),
```
joaodubas: I think it would be better, first to check if this user can perform the action, then hit the database to fetch the development.

apps/re_web/lib/graphql/resolvers/images.ex
```diff
@@ -56,6 +57,16 @@ defmodule ReWeb.Resolvers.Images do
     end)
   end
 
+  def insert_image(%{input: %{parent_type: :development, parent_uuid: parent_uuid} = params}, %{
+        context: %{current_user: current_user}
+      }) do
+    with {:ok, development} <- Developments.get_preloaded(parent_uuid, [:images]),
```
rhnonose: I couldn't find the need for this `preload`, what's the purpose?

rhnonose:
Good job on the `resolve_type` thing, looks awesome.
Did you test with any client? Changing from `non_null` to nullable might break some query typing, you should check with the clients.
This type in particular may not be affected since it's not a top level type, but I'm not sure.

pmargreff:
> Did you test with any client? Changing from non_null to nullable might break some query typing, you should check with the clients.

Didn't saw this possibility, will ask them if it's possible, and in this case, test to try to find integration problems.

apps/re_web/lib/graphql/resolvers/images.ex
rhnonose: My bad, missed that one.

PR #475, opened 2019-04-08
Add query/mutation to handle tags

joaodubas:
Added the following queries:

1. `tags`: list all tags on the system
2. `tag`: fetch a specific tag
3. `tagsSearch`: search tags by ther `name_slug`

Added the following mutations:

1. `tagInsert`: to create a new tag
2. `tagUpdate`: to update details of an existing tag


apps/re_web/lib/graphql/resolvers/tags.ex
```diff
@@ -0,0 +1,48 @@
...
+      end
+
+    {:ok, tags}
+  end
+
+  def search(%{name: name}, %{context: %{current_user: current_user}}) do
+    with :ok <- Bodyguard.permit(Tags, :search, current_user, %{name: name}) do
+      {:ok, Tags.search(name)}
+    else
+      _ -> {:error, :unauthorized}
```
rhnonose: This `else` clause is unnecessary (`Tags.Policy` module returns the same format)

apps/re_web/lib/graphql/resolvers/tags.ex
```diff
@@ -0,0 +1,48 @@
...
+      Tags.get(uuid)
+    else
+      _ -> Tags.get_public(uuid)
+    end
+  end
+
+  def insert(%{input: params}, %{context: %{current_user: current_user}}) do
+    with :ok <- Bodyguard.permit(Tags, :insert, current_user, params),
+         {:ok, tag} <- Tags.insert(params) do
+      {:ok, tag}
```
rhnonose: If the return of the last `with` clause is the return of the `do` block, the last clause should be in the `do` block.

apps/re_web/lib/graphql/resolvers/tags.ex
```diff
@@ -0,0 +1,48 @@
...
+    end
+  end
+
+  def update(%{uuid: uuid, input: params}, %{context: %{current_user: current_user}}) do
+    with :ok <- Bodyguard.permit(Tags, :insert, current_user, params),
+         {:ok, tag} <- Tags.get(uuid),
+         {:ok, tag} <- Tags.update(tag, params) do
+      {:ok, tag}
```
rhnonose: Same as the comment above.

apps/re_web/lib/graphql/resolvers/tags.ex
```diff
@@ -0,0 +1,48 @@
...
+    else
+      _ -> {:error, :unauthorized}
+    end
+  end
+
+  def show(%{uuid: uuid}, %{context: %{current_user: current_user}}) do
+    with :ok <- Bodyguard.permit(Tags, :show, current_user, %{uuid: uuid}) do
+      Tags.get(uuid)
+    else
+      _ -> Tags.get_public(uuid)
```
rhnonose: I'm not sure if this logic should be here in the resolver.
`Bodyguard` should only say if the action is permitted or not (in this case, it is for all users), but the `Tags` module should figure out by the user what content to return.

apps/re_web/lib/graphql/resolvers/tags.ex
```diff
@@ -0,0 +1,48 @@
...
+  Resolvers for tags.
+  """
+  alias Re.Tags
+
+  def index(_params, %{context: %{current_user: current_user}}) do
+    tags =
+      with :ok <- Bodyguard.permit(Tags, :fetch_all, current_user, nil) do
+        Tags.all()
+      else
+        _ -> Tags.public()
```
rhnonose: Same comment about permission/content below.

apps/re_web/test/graphql/tags/query_test.exs
```diff
@@ -0,0 +1,316 @@
...
+      resp_public =
+        post(conn, "/graphql_api", AbsintheHelpers.query_wrapper(query, variable_public))
+
+      assert %{"name" => tag_public.name, "visibility" => "public"} ==
+               json_response(resp_public, 200)["data"]["tag"]
+
+      resp_private =
+        post(conn, "/graphql_api", AbsintheHelpers.query_wrapper(query, variable_private))
+
+      assert nil == json_response(resp_private, 200)["data"]["tag"]
```
rhnonose: This should be `refute` instead of comparing to `nil`

apps/re_web/test/graphql/tags/query_test.exs
```diff
@@ -0,0 +1,316 @@
...
```
rhnonose: This should be `refute` instead of comparing to `nil`

apps/re_web/test/graphql/tags/query_test.exs
```diff
@@ -0,0 +1,316 @@
...
+      ) {
+        tagsSearch(name: $name) {
+          name_slug
+        }
+      }
+      """
+
+      conn = post(conn, "/graphql_api", AbsintheHelpers.query_wrapper(query, variables))
+
+      assert nil == json_response(conn, 200)["data"]["tagsSearch"]
```
rhnonose: This should be `refute` instead of comparing to `nil`

apps/re_web/test/graphql/tags/query_test.exs
```diff
@@ -0,0 +1,316 @@
...
```
rhnonose: This should be `refute` instead of comparing to `nil`

apps/re_web/lib/graphql/resolvers/tags.ex
pmargreff: agreed

joaodubas:
> One thing that caught my eye is the `tagsSearch` query, seems to be a lot of REST-oriented. We can allow this filter directly in tags (index) instead, how we already do on listings, so we don't need to have the same query structure with different capabilities.

I think they solve different needs, `tagsSearch` should be used by an autocomplete system. Besides that, `tagsSearch` can only be acessed by admin users.

> Another implicit gain is, once the filters are implemented in tags, and we add tags on listings we can filter a listing list by tags for free (not the best approach, but good enough to have a proof-of-concept about tags use).

I don't know if it would be that great to filter `listings` by tag like values, such as:

```sql
SELECT
  l.id
FROM
  listings l
  inner join listings_tags lt
    on l.uuid = lt.listing_uuid
  inner join tags t
    on t.uuid = lt.tag_uuid
where
  t.name_slug like '%?'
```

> Was this approach considered while developing? What are your thoughts about it?

I didn't thought about this approach, if it makes sense, I can adjust `index` query to make use of it. But at a first glance it's seems odd to me.

rhnonose:
I think it should have both.
A separate query for autocomplete it totally fine, even being REST is justifiable for caching purposes (if it becomes a bottleneck).
`index` should have a filter as well for category and visibility, for example. Creating the tag filter type has meaning for GraphQL APIs.

pmargreff:
> I think they solve different needs, tagsSearch should be used by an autocomplete system. Besides that, tagsSearch can only be acessed by admin users.
 
Got it, at first I didn't understand the use, I thought the use and the difference from tags' index. 

LGTM :rocket:

apps/re/lib/tags/tags.ex
```diff
@@ -36,6 +36,19 @@ defmodule Re.Tags do
...
+    |> Repo.all()
+  end
+
+  def filter(params, _) do
+    params
+    |> Map.merge(%{visibility: "public"})
+    |> Queries.filter_by()
+    |> Repo.all()
+  end
+
```
rhnonose: These two functions have 3 repeating lines.
It's simpler if it's like this:
```
  def filter(params, user) do
    params
    |> query_visibility(user)
    |> Queries.filter_by()
    |> Repo.all()
end

  defp query_visibility(params, %{role: "admin"}), do: params
  defp query_visibility(params, _), do: Map.merge(params, %{visibility: "public"})
```

PR #476, opened 2019-04-08
Zapier buyer leads

rhnonose:
Zapier integration to save facebook buyer leads


apps/re/lib/leads/facebook_buyer.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Re.Leads.FacebookBuyer do
+  @moduledoc """
+  Schema for facebook buyer leads
+  """
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
+  @primary_key {:uuid, :binary_id, autogenerate: false}
```
pmargreff: Do you think it worts to autogenerate UUIDs once centralized generation is not an requirement anymore?

PR #477, opened 2019-04-11
Refactor zapier integration to validate payload through changeset

rhnonose:
Zapier's integration warned that `neighborhoods` is optional, so I'm adding a proper changeset validation rather than a poorly-written pattern match.

### DIFF
--- a/apps/re/lib/leads/facebook_buyer.ex
+++ b/apps/re/lib/leads/facebook_buyer.ex
@@ -20,7 +20,17 @@ defmodule Re.Leads.FacebookBuyer do
     timestamps()
   end
 
-  @params ~w(full_name email phone_number neighborhoods timestamp lead_id location)a
+  @required ~w(full_name email phone_number timestamp lead_id location)a
+  @optional ~w(neighborhoods)a
+  @params @required ++ @optional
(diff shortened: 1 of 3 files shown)


apps/re/lib/leads/facebook_buyer.ex
```diff
@@ -20,7 +20,17 @@ defmodule Re.Leads.FacebookBuyer do
...
+  @params @required ++ @optional
+  @locations ~w(RJ SP)
 
-  def changeset(struct, params \\ %{}), do: cast(struct, params, @params)
+  def changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, @params)
+    |> validate_required(@required)
+    |> validate_inclusion(:location, @locations,
+      message: "should be one of: [#{Enum.join(@locations, " ")}]"
```
joaodubas: We could add a helper method for this type of validation, I see a lot of choices using it.

rhnonose: I think it could actually be simpler like `message: "should be one of: #{@locations}"`, not sure if it's worth extracting this sort of simplicity.

PR #478, opened 2019-04-11
Allow to show and edit tags in listings queries/mutations

joaodubas:
The last step of the tagging system is to be available in listings, so the following changes were made:

1. Listing type has a `tags` node:
   ```graphql
   query {
     listings {
       listings {
         id
         tags {
           nameSlug
           category
           visibility
         }
       }
     }
   }
   ```
2. Listing input type has a `tags` uuid list:
   ```graphql
   mutation {
     insertListing(input: {
       type: "Apartamento",
       address: {
         state: "SP",
         city: "São Paulo",
         neighborhood: "Vila Pompéia",
         postalCode: "05027-000",
         street: "Rua Doutor Miranda de Azevedo",
         streetNumber: "189",
         lat: -23.5308362,
         lng: -46.6909294
       },
       tags: [
         "3157962a-deac-46e1-a74f-9c9c45e50baf",
         "bc7317e6-480d-4ae4-a16a-ff94e4461499",
         "31ab8cd1-cfe7-4517-af26-7f693141c32d"
       ]
     }) {
       id
       tags {
         nameSlug
       }
     }
   }
   ```

### DIFF
--- a/apps/re/lib/listings/listings.ex
+++ b/apps/re/lib/listings/listings.ex
@@ -234,6 +234,8 @@ defmodule Re.Listings do
     |> Repo.all()
   end
 
+  def upsert_tags(listing, nil), do: {:ok, listing}
+
   def upsert_tags(listing, tag_uuids) do
     tags = Tags.list_by_uuids(tag_uuids)
 
--- a/apps/re/test/listings/listings_test.exs
+++ b/apps/re/test/listings/listings_test.exs
@@ -488,7 +488,7 @@ defmodule Re.ListingsTest do
     end
   end
 
-  describe "upsert_tags/3" do
+  describe "upsert_tags/2" do
     test "should insert tags" do
       tag_1 = insert(:tag, name: "tag 1", name_slug: "tag-1")
       tag_2 = insert(:tag, name: "tag 2", name_slug: "tag-2")
@@ -517,10 +517,7 @@ defmodule Re.ListingsTest do
 
       user = insert(:user)
 
-      {:ok, listing} =
-        insert(:listing, user: user)
-        |> Repo.preload([:tags])
-        |> Listings.upsert_tags([tag_1.uuid, tag_2.uuid])
+      listing = insert(:listing, user: user, tags: [tag_1, tag_2])
 
       {:ok, updated_listing} = Listings.upsert_tags(listing, [tag_3.uuid])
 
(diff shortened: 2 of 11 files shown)


apps/re/lib/listings/listings.ex
```diff
@@ -234,6 +234,8 @@ defmodule Re.Listings do
     |> Repo.all()
   end
 
+  def upsert_tags(listing, nil), do: {:ok, listing}
```
rhnonose: Should also handle empty list

apps/re_web/lib/graphql/types/listing.ex
```diff
@@ -108,6 +110,8 @@ defmodule ReWeb.Types.Listing do
     field :address_id, :id
 
     field :development_uuid, :uuid
+
+    field :tags, list_of(:uuid)
```
rhnonose: This should be `non_null(list_of(non_null(:uuid)))` and a covering test for those cases.

apps/re/lib/listings/listings.ex
joaodubas: An empty list should remove associated tags (https://github.com/emcasa/backend/blob/8edd15ba6bcf56b453dbcbc27e35e09ac4196166/apps/re/test/listings/listings_test.exs#L530-L558)

apps/re_web/lib/graphql/types/listing.ex
joaodubas: Tags can be null (maybe the user don't know what to insert, or the interface is only available for admin users at the launch), and the list can be empty, this way, we can remove all tags from a listing.

joaodubas: Besides that, both cases are covered in tests for `Re.Listings.upsert_tags/2`
https://github.com/emcasa/backend/blob/8edd15ba6bcf56b453dbcbc27e35e09ac4196166/apps/re/test/listings/listings_test.exs#L530-L558

apps/re/test/listings/listings_test.exs
```diff
@@ -536,17 +533,29 @@ defmodule Re.ListingsTest do
...
+      listing = insert(:listing, user: user, tags: [tag_1, tag_2])
 
       {:ok, updated_listing} = Listings.upsert_tags(listing, [])
 
       assert Enum.count(updated_listing.tags) == 0
       refute Enum.member?(updated_listing.tags, tag_1)
       refute Enum.member?(updated_listing.tags, tag_2)
     end
+
+    test "noop when no tags is nil" do
```
pmargreff: Is the test description right? I guess it should be "shouldn't update if receives nil" or something like that ...

apps/re_web/lib/graphql/types/listing.ex
rhnonose: The empty list makes sense, but being able to pass `["someuuid", nil]` doesn't.
So, it should be `list_of(non_null(:uuid))`

rhnonose:
I think that's what I got confused in the empty array comment @pmargreff 
Creating/deleting the `listing <-> tag` relationship sounds better, but it's too SQLy, but having a add/remove mutation could be more explicit.

joaodubas:
I think that having two mutations to handle associate/disassociate tag and listing can be cumbersome for the client, but it's a safer and explicit, will work on this in another PR.

rhnonose:
Yeah, we can discuss it later, it's fine for this PR.

PR #479, opened 2019-04-11
Update development image

pmargreff:
It allows to updated images associated with developments. The major work here was in `check_same_listing` old version only used to look to associated listings. After alter this I extract the logic to a module. I got the "Parents" name from the original GraphQL field `parent_listing` who become only `parent`. 

I also changed the interface from the `check_same_listing` who used to received `[{:ok, images, params}]`, once only images are important to say if they're from the same parent I've extracted only the images from this structure. It'll also enable us to remove duplicated code between [`check_same_listing`](https://github.com/emcasa/backend/blob/master/apps/re/lib/images/images.ex#L101) and  [`fetch_listing`](https://github.com/emcasa/backend/blob/master/apps/re/lib/images/images.ex#L108) which use to contain the same logic for different structures. This deduplication will come in other PR to make the review for this one easier. 

Last but not least, also made a minor change on the public interface: The `distinct_listings` error string is now `distinct_parents`. If the front end use to throw this error, it should be updated in the client. cc: @gabiseabra

### DIFF
--- a/apps/re/lib/images/parents.ex
+++ b/apps/re/lib/images/parents.ex
@@ -8,11 +8,19 @@ defmodule Re.Images.Parents do
     Listings
   }
 
-  def get_parent_from_image_list(images) do
-    cond do
-      unique_listing_parent?(images) -> get_listing_parent(images)
-      unique_development_parent?(images) -> get_development_parent(images)
-      true -> {:error, :distinct_parents}
+  def get_parent_from_image_list([%{listing_id: listing_id} | _] = images)
+      when not is_nil(listing_id) do
+    case unique_listing_parent?(images) do
+      true -> get_listing_parent(images)
+      false -> {:error, :distinct_parents}
+    end
+  end
+
+  def get_parent_from_image_list([%{development_uuid: development_uuid} | _] = images)
+      when not is_nil(development_uuid) do
+    case unique_development_parent?(images) do
+      true -> get_development_parent(images)
+      false -> {:error, :distinct_parents}
     end
   end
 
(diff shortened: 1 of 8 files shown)


apps/re/lib/images/parents.ex
```diff
@@ -0,0 +1,45 @@
...
+    Listings
+  }
+
+  def get_parent_from_image_list(images) do
+    cond do
+      unique_listing_parent?(images) -> get_listing_parent(images)
+      unique_development_parent?(images) -> get_development_parent(images)
+      true -> {:error, :distinct_parents}
+    end
+  end
```
rhnonose: This `cond` is awkward, should be split into 2 public functions and be used inside `Re.Images` with the other suggestion.

apps/re_web/lib/graphql/resolvers/images.ex
```diff
@@ -78,12 +78,18 @@ defmodule ReWeb.Resolvers.Images do
 
   def update_images(%{input: inputs}, %{context: %{current_user: current_user}}) do
     with {:ok, images_and_inputs} <- Images.get_list(inputs),
-         {:ok, listing} <- Images.check_same_listing(images_and_inputs),
-         :ok <- Bodyguard.permit(Images, :update_images, current_user, listing),
+         {:ok, parent} <-
+           images_and_inputs
+           |> extract_image_list()
+           |> Images.Parents.get_parent_from_image_list(),
```
rhnonose: These 3 lines should be inside `Re.Images` wrapped in a  get_parent` function.

PR #480, opened 2019-04-11
Alter deactivation images

pmargreff:
Continuation from #479. It allows deactivating images who has developments as parents. 

It also removes image activation mutation and associated logic, once it seems to be unused. Would you double-check this confirmation on Apollo history also?


pmargreff:
@rhnonose merged the new approach and made the related changes, would you review again?

rhnonose:
It still requires checking if the client uses the REST version or not.
If it does, it shouldn't be removed because it's gonna be migrated at some point.

PR #490, opened 2019-04-15
FB Ads Product Catalog: exporting price_per_area

gabriel-laet:


apps/re/lib/exporters/facebook_ads/product.ex
```diff
@@ -143,6 +143,10 @@ defmodule Re.Exporters.FacebookAds.Product do
     {"custom_label_4", %{}, area}
   end
 
+  defp convert_attribute(:price_per_area, %{price_per_area: price_per_area}) do
```
pmargreff: maybe match :price_per_area against `nil` is a good idea once Float.ceil(nil) will not work.

PR #494, opened 2019-04-16
Buyer lead digestion pipeline for Grupozap Buyer Leads

rhnonose:
- Sets up `ecto_job` to process data asynchronously but keeping consistency
- Digests `grupozap_buyer_leads`, extracting only the relevant information
- Possibly associate `user` by naively looking up phone number
- Possibly associate `listing`


apps/re/lib/leads/job_queue.ex
```diff
@@ -0,0 +1,63 @@
...
+      phone_number: phone_number,
+      origin: gzb.lead_origin,
+      user_uuid: extract_user_uuid(phone_number),
+      listing_uuid: extract_listing_uuid(gzb.client_listing_id)
+    })
+  end
+
+  defp persist(changeset, multi), do: Multi.insert(multi, :insert_buyer_lead, changeset)
+
+  defp extract_user_uuid(phone_number) do
```
pmargreff: This query could be made in `Users` module ...

apps/re/lib/leads/job_queue.ex
```diff
@@ -0,0 +1,63 @@
...
+    case Repo.get_by(User, phone: phone_number) do
+      nil -> nil
+      user -> user.uuid
+    end
+  end
+
+  defp extract_listing_uuid(id) do
+    case Repo.get(Listing, id) do
```
pmargreff: This query could be replaced by `Listings.get`.

PR #497, opened 2019-04-17
Fix job not calling transaction

rhnonose:
Fix `JobQueue` not calling `Repo.transaction/1` when performing.
The mistake was assuming that the library would get the return of the `multi` attribute and perform the call.


rhnonose:
Tested in staging successfully.

PR #498, opened 2019-04-17
Enable district (de)activate

rhnonose:
- Add `is_active` attribute to `districts` so we can keep districts inactive in the database
- Only query active districts


rhnonose:
We should add something like that soon.
For now it should be just to be able to toggle the visibility globally, not regarding user roles.

pmargreff:
My 2 cents: Every time I need an `is_something` or most of the bool fields (not all) I tend to use an `enum/strings` field, it tends to be more future proof in a way it doesn't need to be migrated in case the dimensions of the field need new options. On this case, I would prefer `status`. I know it's very opinionated, but it gives me 0 rework until today now and helps me to save time writing migrations ...

PR #505, opened 2019-04-18
Add command to populate tags

joaodubas:
Add `mix command` to populate tags in system (`mix re.tags.create`). Also add new category for tags.


joaodubas:
I think the problem here, is that [`Ecto.Repo.insert_all/3`](https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert_all/3) doesn't accept a `changeset`.

PR #506, opened 2019-04-22
Do not autogenerate uuids

rhnonose:
To be able to use `ecto_job` better, I'm refactoring to generate uuid in the changeset so we can create a job with the to-be-inserted uuid.


pmargreff:
Good tackle.

PR #510, opened 2019-04-24
Copy units info to listing

pmargreff:
These are necessary changes to replicate units info on listings (so we can use these listings on exporters). The shortcut used is to copy unit params after each time a unit is inserted or updated. Right now I'm using the cheapest unit as reference. 

Highlights are: 
- Copy all unit params instead only the price to parent listing.
- Added test on `unit_update`, to ensure every time we update a unit it will check if the listing copy is up to date. 

Fixed the follow bugs: 
- Initializing units copy server on application. :man_facepalming:  
- Before I used do getting all units on each update (even the inactive ones), now used only the active ones are used. We'll not copy inactive unit info if the cheapest one is inactive.


rhnonose:
Do you think it's too costly to refactor it using `ecto_job`? So we can ensure consistency.

apps/re/lib/listings/units/propagator.ex
```diff
@@ -8,12 +8,25 @@ defmodule Re.Listings.Units.Propagator do
...
+      maintenance_fee: unit.maintenance_fee,
+      floor: unit.floor,
+      rooms: unit.rooms,
+      bathrooms: unit.bathrooms,
+      restrooms: unit.restrooms,
+      area: unit.area,
+      garage_spots: unit.garage_spots,
+      garage_type: unit.garage_type,
+      suites: unit.suites,
+      balconies: unit.balconies
```
joaodubas: Do we have information related to `units_per_floor`, `floor_count`, `elevators`?

apps/re/lib/listings/units/propagator.ex
```diff
@@ -8,12 +8,25 @@ defmodule Re.Listings.Units.Propagator do
...
+    }
```
rhnonose: You can use `Map.take/2` or `Enum.take/2`.

apps/re/lib/listings/units/propagator.ex
pmargreff: I've no idea, I'll ask this to the ~steak~stakeholders.

apps/re/test/listings/units/server_test.exs
```diff
@@ -29,5 +29,24 @@ defmodule Re.Listings.Units.ServerTest do
...
+                 %{
+                   topic: "update_unit",
+                   type: :update,
+                   content: %{
+                     new: unit
+                   }
+                 },
+                 []
+               )
+    end
```
rhnonose: New tests on this `gen_server` + `pub_sub` flow should be tested by spinning up like in [this test](https://github.com/emcasa/backend/blob/979b81cf516829ad28f63ddd6881addbbf76c11a/apps/re/test/listings/listings_test.exs#L311)

apps/re/lib/listings/units/propagator.ex
pmargreff: Yes, we've got this information. 

 I'll introduce it in other PR once it'll required migrations, alter schemas, so on, makes sense to you?

pmargreff:
> Do you think it's too costly to refactor it using ecto_job? So we can ensure consistency.

I have no idea, but I think it's worth invest time, once I'm not 100% confident about `ecto_job` API and it's a good way to get a deeper understanding in the lib. I'll make these changes on other branches once they could accumulate some complexity to the PR. Makes sense to you do it on this way?

rhnonose:
> > Do you think it's too costly to refactor it using ecto_job? So we can ensure consistency.
> 
> I have no idea, but I think it's worth invest time, once I'm not 100% confident about `ecto_job` API and it's a good way to get a deeper understanding in the lib. I'll make these changes on other branches once they could accumulate some complexity to the PR. Makes sense to you do it on this way?

I actually removed the message because we should postpone it, since there's a few `gen_server`s that should be removed in the future.
I posted the message before reviewing, I glanced at the diffs and though it was being introduced now rather than just being spun up.

apps/re/lib/listings/units/propagator.ex
joaodubas: Sounds nice 🤘

apps/re/test/listings/units/server_test.exs
pmargreff: The tests are made on this way in [`units_test`](https://github.com/emcasa/backend/pull/510/files#diff-c52c2b4c749a0d19fa16e6206bd788a5) but I forgot to update them (once the behavior is the same, but extended to other params besides price they didn't break). :man_facepalming: 

Should be ok now, also deleted the files checking only against the interfaces.

apps/re/lib/listings/units/propagator.ex
```diff
@@ -6,26 +6,15 @@ defmodule Re.Listings.Units.Propagator do
...
-      bathrooms: unit.bathrooms,
-      restrooms: unit.restrooms,
-      area: unit.area,
-      garage_spots: unit.garage_spots,
-      garage_type: unit.garage_type,
-      suites: unit.suites,
-      balconies: unit.balconies
-    }
+    params =
+      Enum.min_by(units, fn unit -> Map.get(unit, :price) end)
```
rhnonose: This should be:
```elixir
units
|> Enum.min...
|> Map.take...
```
AFAIK it's a credo warning, Codacy should've caught that.

PR #511, opened 2019-04-25
Add owner contact information for listing

joaodubas:
We were missing one information from the listing, who is the seller and how can we contact him/her.

From now on, a listing can be associated with a owner contact, that has all contact information (phone and email).


apps/re/lib/listings/schemas/listing.ex
```diff
@@ -53,6 +53,11 @@ defmodule Re.Listing do
 
     belongs_to :user, Re.User
 
+    belongs_to :owner, Re.OwnerContact,
```
pmargreff: Why not call it `owner_contact`? It keeps the pattern to use the table name the same in association name.

rhnonose: also, graphql's schema already has `owner` for the associated user, which may conflict.

joaodubas: 🆗 renaming it!

PR #512, opened 2019-04-25
feat(seller_leads): add seller lead schema for the site

rhnonose:
I'm adding a seller lead schema to decouple the leads that currently gets placed directly as a listing. This should make it clear what's a lead and what's an inactive listing and enable better control over our listing management


apps/re/lib/leads/site_seller.ex
```diff
@@ -0,0 +1,50 @@
...
+  import Ecto.Changeset
+
+  @primary_key {:uuid, :binary_id, autogenerate: false}
+
+  schema "site_seller_leads" do
+    field :complement, :string
+    field :type, :string
+    field :maintenance_fee, :float
+    field :suites, :integer
+    field :price, :integer
```
joaodubas: Is this all information that we get? I thought we asked:

1. address
2. bedrooms
3. bathrooms

rhnonose: So, this is a record to save the lead coming from the site.
At this point, all those informations are saved either through `tour_appointment` or `price_suggestion`, so we can retrieve in the next step when processing a table for the integration.

PR #513, opened 2019-04-25
Seller lead mutation

rhnonose:
Creating a new mutation for inserting seller leads so it doesn't go directly to `listings`.
`price_request` and `tour_appointment` is being added here to keep additional information and keep track.


apps/re/lib/seller_leads/site_lead.ex
```diff
@@ -0,0 +1,50 @@
...
+    field :suites, :integer
+    field :price, :integer
+
+    belongs_to :price_request, Re.PriceSuggestions.Request
+    belongs_to :tour_appointment, Re.Calendars.TourAppointment
+
+    timestamps()
+  end
+
+  @types ~w(Apartamento Casa Cobertura)
```
pmargreff: It's a good time to translate these values into English.

rhnonose: Hmm, this will require a mapping in the frontend that uses the same structure to insert listings, so it might generate too much effort for the time being.
Thoughts @gabiseabra ?

apps/re/lib/seller_leads/site_lead.ex
```diff
@@ -0,0 +1,50 @@
...
+    field :complement, :string
+    field :type, :string
+    field :maintenance_fee, :float
+    field :suites, :integer
+    field :price, :integer
+
+    belongs_to :price_request, Re.PriceSuggestions.Request
+    belongs_to :tour_appointment, Re.Calendars.TourAppointment
+
+    timestamps()
```
joaodubas: We could start using the timestamp with timezone for new schemas.

apps/re_web/test/graphql/seller_leads/mutation_test.exs
```diff
@@ -0,0 +1,125 @@
...
+
+      assert seller_lead["uuid"]
+      assert seller_lead["complement"] == "100"
+      assert seller_lead["type"] == "Apartamento"
+      assert seller_lead["price"] == 800_000
+      assert seller_lead["maintenanceFee"] == 100.00
+      assert seller_lead["suites"] == 2
+    end
+
+    test "anonymous should add seller lead", %{
```
joaodubas: This test should be named : `anonymous user should not add seller lead`.

apps/re/lib/seller_leads/site_lead.ex
```diff
@@ -0,0 +1,50 @@
...
+  @moduledoc """
+  Schema for site's seller leads
+  """
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
+  @primary_key {:uuid, :binary_id, autogenerate: false}
+
+  schema "site_seller_leads" do
```
joaodubas: I think the migration for this schema is missing.

joaodubas: Forget about this, I found it in a previous PR: https://github.com/emcasa/backend/blob/04c758ebb65389799f61616bed9f8cc55c2e4c80/apps/re/priv/repo/migrations/20190424180924_create_site_seller_leads.exs

rhnonose: What the fuck?

rhnonose: Ahh, I thought it was being ignored or something 😅

apps/re/lib/seller_leads/site_lead.ex
rhnonose: @lucas-tulio

apps/re/lib/seller_leads/site_lead.ex
```diff
@@ -0,0 +1,50 @@
...
+    field :complement, :string
+    field :type, :string
+    field :maintenance_fee, :float
+    field :suites, :integer
+    field :price, :integer
+
+    belongs_to :price_request, Re.PriceSuggestions.Request
+    belongs_to :tour_appointment, Re.Calendars.TourAppointment
+
+    timestamps(type: :utc_datetime)
```
joaodubas: I think the migration should also be adjusted so both timestamps are kept with timezone info, check this migration:
https://github.com/emcasa/backend/blob/4f559f52c4d074888bbbc97d7977d88f1a5b0c3b/apps/re/priv/repo/migrations/20190423170422_add_owner_contact.exs#L14

apps/re/lib/seller_leads/site_lead.ex
lucas-tulio: @rhnonose I think it's a good idea to map these values. We just have to sync it with a frontend release because it would make the current mutation incompatible.

PR #515, opened 2019-04-25
Add infrastructure attributes to developments

pmargreff:
Add new infrastructure attributes on developments. They're softly duplicated on listings by now duo the filters and exporters.

Made a name type from `unit_per_floor` (listings) to `units_per_floor` on developments, do you think whort it to rename it on listings as well? 

Also add GraphQL entries for it on development cc: @gabiseabra.

### DIFF
--- a/apps/re/lib/developments/development.ex
+++ b/apps/re/lib/developments/development.ex
@@ -15,6 +15,9 @@ defmodule Re.Development do
     field :phase, :string
     field :builder, :string
     field :description, :string
+    field :floor_count, :integer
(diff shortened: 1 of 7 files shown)


joaodubas:
Maybe we could rename both:

* `s/unit_per_floor/units_per_floor`
* `s/floor_count/floors`

In both cases I didn't follow the naming convention adopted in other fields.

pmargreff:
@joaodubas makes sense, I'll submit these changes in other PR.

PR #517, opened 2019-04-30
Filter development listings

pmargreff:
Allow the client to filter listings while querying from developments as root. It doesn't accept pagination. 

Also made some minor refactor extracting `has_admin_rights` to Listings context once it is used in different places and rewrite developments `has_admin_rights` to look only to user role (once only admin has admin rights on this contexts).

### DIFF
--- a/apps/re/test/listings/listings_test.exs
+++ b/apps/re/test/listings/listings_test.exs
@@ -695,6 +695,27 @@ defmodule Re.ListingsTest do
     end
   end
 
+  describe "has_admin_rights?/2" do
+    test "is true when user is admin" do
+      listing = build(:listing)
+      user = %{role: "admin"}
+      assert Listings.has_admin_rights?(listing, user) == true
+    end
+
+    test "is true when user is owner" do
+      user = insert(:user)
+      listing = insert(:listing, user: user)
+      assert Listings.has_admin_rights?(listing, user) == true
+    end
+
+    test "is false if user isn't owner neither admin" do
+      owner = insert(:user)
+      other_user = insert(:user, role: "user")
+      listing = insert(:listing, user: owner)
+      assert Listings.has_admin_rights?(listing, other_user) == false
+    end
+  end
+
   defp chunk_and_short(listings) do
     listings
     |> Enum.chunk_by(& &1.score)
--- a/apps/re/test/listings/listings_test.exs
+++ b/apps/re/test/listings/listings_test.exs
@@ -695,27 +695,6 @@ defmodule Re.ListingsTest do
     end
   end
 
-  describe "has_admin_rights?/2" do
-    test "is true when user is admin" do
-      listing = build(:listing)
-      user = %{role: "admin"}
-      assert Listings.has_admin_rights?(listing, user) == true
-    end
-
-    test "is true when user is owner" do
-      user = insert(:user)
-      listing = insert(:listing, user: user)
-      assert Listings.has_admin_rights?(listing, user) == true
-    end
-
-    test "is false if user isn't owner neither admin" do
-      owner = insert(:user)
-      other_user = insert(:user, role: "user")
-      listing = insert(:listing, user: owner)
-      assert Listings.has_admin_rights?(listing, other_user) == false
-    end
-  end
-
   defp chunk_and_short(listings) do
     listings
     |> Enum.chunk_by(& &1.score)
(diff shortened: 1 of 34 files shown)


rhnonose:
LGTM 👍 
But I think if we're refactoring `has_admin_rights?`, we should put in context's `Policy` module.
This was a lazy workaround because our `Policy` logic didn't return a boolean, but permissions should be there.

apps/re/test/listings/listings_test.exs
```diff
@@ -695,6 +695,27 @@ defmodule Re.ListingsTest do
...
+  describe "has_admin_rights?/2" do
+    test "is true when user is admin" do
+      listing = build(:listing)
+      user = %{role: "admin"}
+      assert Listings.has_admin_rights?(listing, user) == true
+    end
+
+    test "is true when user is owner" do
+      user = insert(:user)
+      listing = insert(:listing, user: user)
```
joaodubas: Maybe you can just populate the Listing and the User:

```elixir
user = Re.User{id: 1}
listing = Re.Listing{user_id: 1}
```

joaodubas: Maybe you can just populate the Listing and the User:

```elixir
user = Re.User{id: 1}
listing = Re.Listing{user_id: 1}
```

pmargreff:
> But I think if we're refactoring has_admin_rights?, we should put in context's Policy module. 

Makes sense, I'll change it.

apps/re_web/lib/graphql/resolvers/addresses.ex
```diff
@@ -6,19 +6,26 @@ defmodule ReWeb.Resolvers.Addresses do
...
-    admin? = admin?(current_user)
+  def per_development(development, params, %{context: %{current_user: current_user}}) do
+    admin = access_complete_address?(current_user, development)
 
-    {:address, Map.put(params, :has_admin_rights, admin?)}
+    {:address, Map.put(params, :has_admin_rights, admin)}
+  end
+
+  defp access_complete_address?(user, parent) do
+    case Policy.authorize(:show_complete_address, user, parent) do
```
rhnonose: You shouldn't call `Policy` directly, should should go through the context.

apps/re_web/lib/graphql/resolvers/images.ex
```diff
@@ -121,8 +123,12 @@ defmodule ReWeb.Resolvers.Images do
   def insert_image_trigger(%{parent: %Re.Development{uuid: uuid}}),
     do: "development_updated:#{uuid}"
 
-  defp admin?(%{role: "admin"}), do: true
-  defp admin?(_), do: false
+  defp access_inactive_images?(user, parent) do
+    case Policy.authorize(:show_inactive_images, user, parent) do
```
rhnonose: You shouldn't call Policy directly, should should go through the context.

apps/re_web/lib/graphql/resolvers/addresses.ex
pmargreff: yes, for some reason the version you reviewed is outdated, can you go to the last one?

apps/re_web/lib/graphql/resolvers/images.ex
```diff
@@ -121,16 +123,26 @@ defmodule ReWeb.Resolvers.Images do
   def insert_image_trigger(%{parent: %Re.Development{uuid: uuid}}),
     do: "development_updated:#{uuid}"
 
+  defp has_admin_rights?(user, %Listing{} = listing) do
+    case Listings.authorize(:has_admin_rights, user, listing) do
```
rhnonose: My apologies, I wasn't clear with this one.
It should be used like described in `bodyguard`.
Example: `Bodyguard.permit(Listings, :edit_listing, user, listing)`.
It uses the `Bodyguard.Policy` behaviour.

pmargreff: I got you now, I didn't know it is the right way to do it. 
Thank you for the explanation.

pmargreff: Also [saw in the documentation ](https://hexdocs.pm/bodyguard/Bodyguard.html#permit?/4)that `permit?` does convert the response to bool (:ok -> true, _ -> false) for free.

rhnonose: That's a great finding, didn't know that 🤔

PR #519, opened 2019-04-30
Change scheduleTour mutation to accept seller lead

rhnonose:
This might break the frontend as it puts another parameter to tour appointment

### DIFF
--- a/apps/re_web/test/graphql/calendars/mutation_test.exs
+++ b/apps/re_web/test/graphql/calendars/mutation_test.exs
@@ -25,6 +25,7 @@ defmodule ReWeb.GraphQL.Calendars.MutationTest do
 
   test "admin should schedule tour appointment", %{admin_conn: conn, admin_user: user} do
     listing = insert(:listing)
+    site_seller_lead = insert(:site_seller_lead)
 
     args = %{
       "input" => %{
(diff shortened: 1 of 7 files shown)


apps/re_web/test/graphql/calendars/mutation_test.exs
```diff
@@ -33,7 +34,8 @@ defmodule ReWeb.GraphQL.Calendars.MutationTest do
         ],
         "wantsPictures" => true,
         "wantsTour" => true,
-        "listingId" => listing.id
+        "listingId" => listing.id,
+        "siteSellerLeadUuid" => site_seller_lead.uuid
```
pmargreff: Tour appointment will receive both `listingId` and `siteSellerLeadUUid` at once or only the `siteSellerLeadUuid` and we keep accept `listingId` to not brake the interface? If it's we will receive only `site..` we could create a test case receiving the `siteSellerLeadUuid` but not the `listing_id`, it can avoid a false positive because it is receiving `listingId`.

rhnonose: Yeah, I should split the test.
I'm keeping `listing_id` as a parameter because the mobile app might still call it.

PR #522, opened 2019-05-02
Send notification e-mail for new site seller lead

rhnonose:
Send e-mail in the old `pubsub` format since we'll remove all at once when we don't need the e-mails anymore

### DIFF
--- a/apps/re_integrations/lib/notifications/emails/user.ex
+++ b/apps/re_integrations/lib/notifications/emails/user.ex
@@ -108,6 +108,37 @@ defmodule ReIntegrations.Notifications.Emails.User do
                   #{changes_txt}")
   end
 
+  def new_site_seller_lead(
+        %{price_request: %{user: user, address: address} = price_request} = site_lead
+      ) do
+    new()
+    |> to(@to)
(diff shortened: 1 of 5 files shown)


apps/re_integrations/lib/notifications/emails/user.ex
```diff
@@ -108,6 +108,37 @@ defmodule ReIntegrations.Notifications.Emails.User do
...
+    |> subject("Novo seller lead")
+    |> html_body("Nome: #{user.name}<br>
+                  Email: #{user.email}<br>
+                  Telefone: #{user.phone}<br>
+                  Área: #{price_request.area}<br>
+                  Quartos: #{price_request.rooms}<br>
+                  Banheiros: #{price_request.bathrooms}<br>
+                  Vagas: #{price_request.garage_spots}<br>
+                  Rua: #{address.street}<br>
+                  Número: #{address.street_number}<br>
```
joaodubas: We could send more details about the address.

rhnonose: Working on it.

PR #523, opened 2019-05-02
Put more information on e-mail subject

rhnonose:


apps/re_integrations/lib/notifications/emails/user.ex
```diff
@@ -114,7 +114,7 @@ defmodule ReIntegrations.Notifications.Emails.User do
     new()
     |> to(@to)
     |> from(@admin_email)
-    |> subject("Novo seller lead")
+    |> subject("Novo seller lead de #{address.neighborhood} em #{address.state}")
```
joaodubas: You could add the city.

PR #527, opened 2019-05-03
refactor(stats-counter): Doesn't load structures to count

pmargreff:
This PR improves the performance on dashboard stats counters. The original version used to fetch all entities to memory and count them in language level. This implementation helped us to prevent N+1 queries but overload memory, sometimes causing a timeout on connections 'cause IO time used to dominate. 

The reason for timeout is because sometimes we hit tables with a lot of registries (from the samples I [got on sentry](https://sentry.io/organizations/emcasa/issues/942395465/events/db26e856000d4cf29cc8ecfb989a974c/?project=1389074)) we use to fetch between 65k ~ 100k results at once. This also can be the reason for our memory use to spin up sometimes without any direct correlation with traffic. 

I replace the old solution with a [run_batch](https://hexdocs.pm/dataloader/Dataloader.Ecto.html#module-custom-batch-queries) and postgres `count(*)` plus `group_by`, as we still counting a lot of entries, it is faster but not in magnitude order, on local tests the new query time is 3s ~ 5s where the old one used to be 3s ~ 11s. But we have two other gains, the memory footprint is better once we don't load the schemas in memory, and consequently, cache works great, once we have postgres results in memory. The result for the slowest case (6s~) uses to come in 600ms ~ 800ms. For some reason, the old solution doesn't cache it, probably because the old one was too big that it doesn't fit in memory page.

Last but not least, this query seems to be triggered in garagem search, and 'cause we don't show this info on the search page, only inside listings. I'm taking the opportunity to review and this query from search. @gabiseabra suggestions about where start to look at it are welcome.


apps/re/lib/statistics/dataloader_queries.ex
```diff
@@ -0,0 +1,27 @@
...
+
+  def run_batch(_, query, :count, listings, repo_opts) do
+    listing_ids = Enum.map(listings, & &1.id)
+    default_count = 0
+
+    result =
+      query
+      |> where([p], p.listing_id in ^listing_ids)
+      |> group_by([p], p.listing_id)
+      |> select([p], {p.listing_id, count("*")})
```
joaodubas: Maybe we could use `count(1)` instead of `count(*)`.

```suggestion
      |> select([p], {p.listing_id, count(1)})
```

PR #531, opened 2019-05-06
feat(seller_leads): add pompeia as covered neighborhood

rhnonose:
Sometimes, Google Maps API sends "Vila Pompéia" as "Pompeia". Including "Pompeia" as an accepted neighborhood for now until we fix this address discrepancy.

### DIFF
--- a/apps/re/lib/addresses/neighborhoods.ex
+++ b/apps/re/lib/addresses/neighborhoods.ex
@@ -82,6 +82,7 @@ defmodule Re.Addresses.Neighborhoods do
     %{state: "RJ", neighborhood: "Urca", city: "Rio de Janeiro"},
     %{state: "SP", neighborhood: "Perdizes", city: "São Paulo"},
     %{state: "SP", neighborhood: "Vila Pompéia", city: "São Paulo"},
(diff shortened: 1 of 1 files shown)


apps/re/lib/addresses/neighborhoods.ex
```diff
@@ -82,6 +82,7 @@ defmodule Re.Addresses.Neighborhoods do
     %{state: "RJ", neighborhood: "Urca", city: "Rio de Janeiro"},
     %{state: "SP", neighborhood: "Perdizes", city: "São Paulo"},
     %{state: "SP", neighborhood: "Vila Pompéia", city: "São Paulo"},
+    %{state: "SP", neighborhood: "Pompeia", city: "São Paulo"},
```
pmargreff: Shouldn't be Pompéia instead? `is_covered?` fn know how to normalize it ...

rhnonose: So, the raw form that Google API sends is `Pompeia`, hence the exact "replication".

rhnonose:
@pmargreff this is the only one that's recurring.

PR #533, opened 2019-05-07
Process interests as buyer leads

rhnonose:


apps/re/test/buyer_leads/job_queue_test.exs
```diff
@@ -189,4 +189,43 @@ defmodule Re.BuyerLeads.JobQueueTest do
...
+      %{uuid: uuid} = insert(:interest, listing: listing, phone: "+5511999999999")
+
+      assert {:ok, _} = JobQueue.perform(Multi.new(), %{"type" => "interest", "uuid" => uuid})
+
+      assert buyer = Repo.one(BuyerLead)
+      assert buyer.user_uuid == user_uuid
+      assert buyer.listing_uuid == listing_uuid
+    end
+
+    test "process lead with nil phone" do
```
pmargreff: Does this could happen on users who have no phone but still logged in?

rhnonose: No, but I found some records in the database with `nil` phones, so if we decide to process past records, could raise some errors.

PR #535, opened 2019-05-08
Create a copy field for tour appointment's datetime option

rhnonose:
To be able to be more easily picked up by skyvia, I'm creating a datetime field with a copy of the option being sent. Since the client is opting to send a single option, this will create a copy to be picked up until we are able to change the API (which is a breaking change).

### DIFF
--- a/apps/re/lib/calendars/calendars.ex
+++ b/apps/re/lib/calendars/calendars.ex
@@ -12,12 +12,18 @@ defmodule Re.Calendars do
   defdelegate authorize(action, user, params), to: __MODULE__.Policy
 
   def schedule_tour(params) do
+    option = get_one_datetimte(params)
+
     %TourAppointment{}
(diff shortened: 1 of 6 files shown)


apps/re/lib/calendars/calendars.ex
```diff
@@ -12,13 +12,19 @@ defmodule Re.Calendars do
...
+
     %TourAppointment{}
-    |> TourAppointment.changeset(params)
+    |> TourAppointment.changeset(Map.merge(%{option: option}, params))
     |> add_listing_id(params)
     |> Repo.insert()
     |> PubSub.publish_new("tour_appointment")
   end
 
+  defp get_one_datetimte(%{options: [%{datetime: datetime} | _rest]}), do: datetime
```
pmargreff: type: `defp get_one_datetime...`

PR #543, opened 2019-05-10
Allow to update user role

pmargreff:
Allow admins to update other user roles through mutation. 

Some points: 
- Use the pattern you've suggested earlier, pick `userUpdateRole` to mutation name, so it's sorted by type and action and will be grouped with other similar mutations as `userUpdate`, `userUpdateEmail`... 
- Use the `uuid` on mutation, so it explicit suggest all other actions on users to use the uuid as the identifier.
- Choose to go with update role instead promote to admin, so we could easily extend it when having more roles.    

@beligante take a look when you have some time, it's the other part for the feature we've started, a little bit shorter and better contained, so it would be easier to read and get some food for thought.


pmargreff:
@rhnonose Agreed with the first point, I fell the same. 

About the second one, one long term I infer the following: Supposing we have `user`, `ops`, `sales` and `admin` as roles, only admin can update other users (include other admins). I probably should have discussed instead of supposing :grimacing:. I also agree with your point about granularity on permissions, I'll refactor it, maybe using a state machine to solve this, do you think it's a good idea to introduce a new dependency for it or it's too early?

rhnonose:
@pmargreff nah, just keeping the action `promote_to_admin` is enough for me.
I think neither case should be implemented (generic role update or state machine) for now, just keep it simple until we find out the actual needs so we don't assume things and create the wrong abstraction.

PR #544, opened 2019-05-10
List user info

beligante:


pmargreff:
Allow admins to fetch users. Started to write pagination by hand but in the end, I saw we already use screenever as a dependency (cc: @beligante, @gabiseabra). 

Will submit one more pull request with a search for users.

apps/re_web/lib/graphql/resolvers/accounts.ex
```diff
@@ -78,6 +78,14 @@ defmodule ReWeb.Resolvers.Accounts do
     end
   end
 
+  def users(params, %{context: %{current_user: %{role: "admin"}}}) do
```
rhnonose: You should use a `Policy` module with `Bodyguard` to keep this permission.

apps/re_web/lib/graphql/types/user.ex
```diff
@@ -47,6 +47,19 @@ defmodule ReWeb.Types.User do
     field :app, :boolean
   end
 
+  input_object :pagination_input do
+    field :page, :integer
+    field :page_size, :integer
```
rhnonose: These fields should be required.

PR #547, opened 2019-05-13
Refactor neighborhood

rhnonose:
- Remove hard-coded covered district lookup in favor of database lookup
- Get districts by name/city/state slugs rather than raw values
- In the `is_covered` function, handle address and raw input transparently

This modification will allow to activate/deactive coverage through the database, so we can build a UI for it.
It also prompted me to think about refactoring `Addresses` and `Neighborhoods` which should be one single context (something like `Location`), a refactoring task was created for that.

### DIFF
--- a/apps/re/test/support/factory.ex
+++ b/apps/re/test/support/factory.ex
@@ -113,7 +113,7 @@ defmodule Re.Factory do
     }
   end
 
-  def from_address(address) do
+  def district_from_address(address) do
     insert(:district,
       name: address.neighborhood,
       name_slug: address.neighborhood_slug,
@@ -121,6 +121,7 @@ defmodule Re.Factory do
       city_slug: address.city_slug,
       state: address.state,
       state_slug: address.state_slug,
+      description: Shakespeare.hamlet(),
       status: "active"
     )
   end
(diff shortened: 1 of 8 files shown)


apps/re/test/support/factory.ex
```diff
@@ -93,20 +93,38 @@ defmodule Re.Factory do
...
+      city: city,
+      city_slug: city_slug,
+      state: state,
+      state_slug: state_slug,
       description: Shakespeare.hamlet(),
       status: "active"
     }
   end
 
+  def from_address(address) do
```
pmargreff: This factory could be named as `district_from_address`.

apps/re_web/test/graphql/listings/query_test.exs
```diff
@@ -1103,9 +1103,9 @@ defmodule ReWeb.GraphQL.Listings.QueryTest do
         address = insert(:address)
 
       insert(:district,
-        state: address.state,
-        city: address.city,
-        name: address.neighborhood,
+        state_slug: address.state_slug,
```
pmargreff: If I got right you could use the factory `from_address` in this case and all the above repetitions.

PR #549, opened 2019-05-14
Add update user role to graphql

pmargreff:
The same subject from #543 came back. @gabiseabra point out it's not a good idea have a no reversible action on API, as well add a mutation to each role transition while we are adding roles is a backend dependency which we could avoid. 

We should take care of role update on future using some kind of state machine and it would be transparent to frontend clients. By now, every admin can update all users roles.


rhnonose:
If the problem is a no reversible action, we can pair up with a `downgrade_to_user` mutation.
We should keep the update as simple as we can since it's only two roles. Creating an abstraction for the future is not good, state machine is complicating too much.

PR #550, opened 2019-05-14
fix(user-management): change user list response time

pmargreff:
* Alter response type from userPagination list to userPagination.
* Return user authorization response on query body when user has
no access to user list.
* Add authorization response asserts on user list access tests.


pmargreff:
I guess commit message is descriptive, but any question or unexplained changed feel free to send your questions.

pmargreff:
cc: @gabiseabra

PR #553, opened 2019-05-14
Export tags in vivareal integration

rhnonose:
Export some tags for vivareal integration.
Features mappings (loosely) according to [the documentation](https://sites.google.com/grupozap.com/manualdeintegracaovivareal/integra%C3%A7%C3%A3o-de-im%C3%B3veis/detalhes-do-im%C3%B3vel/benef%C3%ADcios).
Order by `name_slug` just for consistency (`desc` since `Enum.reduce/3` builds the list backwards and `reverse` is costly)

This is the remaining modifications mentioned in #548

@joaodubas not sure if there's any tag mappings missing, since it seems like you got some from them, so please take a look.

### DIFF
--- a/apps/re/lib/exporters/vivareal.ex
+++ b/apps/re/lib/exporters/vivareal.ex
@@ -92,7 +92,8 @@ defmodule Re.Exporters.Vivareal do
     {"Media", %{}, Enum.map(images, &build_image/1)}
   end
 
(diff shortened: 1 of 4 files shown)


rhnonose:
@pmargreff what do you mean by logging? We have logs for when the route is called.

PR #554, opened 2019-05-14
Filter users

pmargreff:
The main idea here is to allow the user to make a simple search on users when changing roles. It's not a full-text search (and it doesn't intend to be), it's only to avoid users to endless navigation, we can refine the way we do with ops' feedbacks. 

I changed how we deal with filtering, I'm also refactoring the ones which already exists (in another branch). Before they use to exist by themselves, now they live inside context dir.  

@gabiseabra about the GraphQL interface, feel free to feedback if have suggestions. I guess we haven't text search yet, this one could help how to make guide similar situations.


rhnonose:
So, seems like you copied `Filtering` from listings, but the logic being implemented doesn't look like filtering but a sort of text search, so the module name should reflect that.
Also, a schema is not necessary for that either, so it should just be regular query module with text search.

pmargreff:
Got it, at the beginning I was with thoughts about the same point but follow the public interface which defines GraphQL filters as internal filter modules :x: . 

I'm changing the implementation but will keep the same public API for frontend.

rhnonose:
You should change the API.
Search and filter can coexist, but this one is search and not filtering because it's more loose.
You're giving a generic field, so there's no specific attribute for filtering.

apps/re/lib/accounts/accounts.ex
```diff
@@ -13,7 +14,9 @@ defmodule Re.Accounts do
   def query(query, _args), do: query
 
   def paginated(params \\ %{}) do
-    Repo.paginate(Re.User, params)
+    User
+    |> Queries.search_by_name_or_email_or_phone(params)
```
rhnonose: You can build `or_where` clauses separately, so this function can be split into 3.

apps/re/lib/accounts/queries.ex
```diff
@@ -0,0 +1,18 @@
...
+  @moduledoc """
+  Module for grouping accounts queries
+  """
+  use Ecto.Schema
+
+  import Ecto.Query
+
+  def search_by_name_or_email_or_phone(query, %{search: search}) do
+    from(u in query,
+      where: like(u.phone, ^"%#{search}%"),
```
rhnonose: Did you try `ilike` for this? Since our numbers are stored with country code, we would need to every time type `+55` to search a number.

pmargreff: Ilike is used to normalize string case, Postgres `ILIKE` match words even when they contain different cases, when dealing with numbers it doesn't improve search. The part responsible by the substring part is `%#{word}%`. So we already don't need to send all string to find it.

rhnonose: My bad, I mistook for something else.

PR #561, opened 2019-05-20
Extract development listings

pmargreff:
As #560 I'm also cherry picking these changes, to better readability. 

In these changes, we split the context to handle development's listings. Now the rules for these listings will live under `developments` context while common listings will continue under `listings` responsibility. Also, remove user reference for developments' listings, once they're not relevant to this situation.


pmargreff:
> Maybe we should use this PR to also rename Propagator to Mirror as we discussed.

This one was removed, I used the context to make these replications. Rename it doesn't make sense anymore. 

I broke the PR like this because it shouldn't change interfaces and we can run this part in prod to get any possible side effects to refactor only ...

PR #565, opened 2019-05-21
Add async listing from new unit

pmargreff:
When a user adds new unit it auto-generate new listing from it, this part will be helpful while integrate to get units from third parts and keep all our model around existing (salesforce, fb, vivareal, so on) integrations working. 

It also removes the propagator module who was I kinda responsible for the same job done on `developments/listings` module is doing (extract info from unit/development). 

It doesn't implement update mirror by now, I'll tackle this part on the nearby future.


apps/re/lib/developments/units/units.ex
```diff
@@ -36,29 +41,50 @@ defmodule Re.Units do
...
+        |> Repo.transaction()
+
+      %{errors: errors} = changeset ->
+        Logger.warn("Invalid payload from new_unit. Errors: #{Kernel.inspect(errors)}")
+
+        {:error, changeset}
+    end
+  end
+
+  def update(unit, params, development, listing \\ nil)
```
pmargreff: this one could be refactored to the accept params in the same way we've done with listings, I created a new card for this task.

apps/re/lib/developments/listings.ex
```diff
@@ -17,6 +20,18 @@ defmodule Re.Developments.Listings do
     |> Repo.insert()
   end
 
+  def multi_insert(multi, params, opts) do
```
pmargreff: Another possible interface is only to add a new action to multi and return multi itself, so we can aggregate operations from different contexts, and roll up the transaction outside this function (in this case on job perform).

apps/re/lib/developments/job_queue.ex
```diff
@@ -0,0 +1,35 @@
...
+    Developments.Listings,
+    Units
+  }
+
+  alias Ecto.Multi
+
+  def perform(%Multi{} = multi, %{"type" => "new_unit", "uuid" => uuid}) do
+    {:ok, %{development_uuid: development_uuid}} = {:ok, unit} = Units.get(uuid)
+    {:ok, %{address_id: address_id}} = {:ok, development} = Developments.get(development_uuid)
+    {:ok, address} = Addresses.get_by_id(address_id)
```
rhnonose: You can put this in one query:
```elixir
Unit
|> Ecto.Query.preload(development: [:address])
|> Repo.get(uuid)
```

apps/re/lib/developments/listings.ex
rhnonose: When the occasion arrives, we can refactor it. It should be fine like this.

apps/re/lib/developments/listings.ex
```diff
@@ -38,6 +53,40 @@ defmodule Re.Developments.Listings do
...
+
+  def listing_params_from_unit(%Re.Unit{} = unit, %Re.Development{} = development) do
+    params_from_unit = extract_listing_params_from_unit(unit)
+    params_from_development = extract_listing_params_from_development(development)
+
+    %{}
+    |> Map.merge(params_from_unit)
+    |> Map.merge(params_from_development)
+    |> Map.merge(@static_attributes)
+  end
```
rhnonose: This function could be more readable if, instead of repeating 3 `Map.merge/2` functions, you could do the merge inside `extract_listing_params_from_unit` and `extract_listing_params_from_development` (and renaming accordingly).

apps/re/lib/developments/listings.ex
```diff
@@ -38,6 +53,40 @@ defmodule Re.Developments.Listings do
...
+    type: "Apartamento",
+    is_release: true,
+    garage_type: "unknown"
+  }
+
+  def listing_params_from_unit(%Re.Unit{} = unit, %Re.Development{} = development) do
+    params_from_unit = extract_listing_params_from_unit(unit)
+    params_from_development = extract_listing_params_from_development(development)
+
+    %{}
```
rhnonose: You can start the pipeline with `@static_attributes`.

apps/re/lib/developments/job_queue.ex
```diff
@@ -0,0 +1,35 @@
...
+  alias Re.{
+    Addresses,
+    Developments,
+    Developments.Listings,
+    Units
+  }
+
+  alias Ecto.Multi
+
+  def perform(%Multi{} = multi, %{"type" => "new_unit", "uuid" => uuid}) do
```
rhnonose: Now that I'm thinking about it, `type` should be the job type instead of an "event", so this would read better as `mirror_unit_to_listing`.

apps/re/lib/developments/job_queue.ex
```diff
@@ -0,0 +1,35 @@
+defmodule Re.Developments.JobQueue do
+  @moduledoc """
+  Module for processing developments listings to extract attributes from unit and
+  create an listing from
```
rhnonose: Moduledoc should be more generic since we just defined as a generic job for developments.
Something like "this is a job queue for processing development stuff, here is one of the actions: extract from unit etc".

apps/re/lib/developments/job_queue.ex
```diff
@@ -0,0 +1,35 @@
...
+    {:ok, %{address_id: address_id}} = {:ok, development} = Developments.get(development_uuid)
+    {:ok, address} = Addresses.get_by_id(address_id)
+
+    params = Listings.listing_params_from_unit(unit, development)
+
+    multi =
+      {:ok, %{listing: listing}} =
+      Listings.multi_insert(multi, params, development: development, address: address)
+
+    Units.update(unit, %{}, development, listing)
```
rhnonose: This update should be part of the `multi` as well because `ecto_job` have retry logic.
If the first `multi_insert` works but this line doesn't, there's gonna be a repeated write.

apps/re/lib/developments/listings.ex
pmargreff: it does make sense :ok_hand:

PR #569, opened 2019-05-21
Enable PostGIS

gabriel-laet:


config/config.exs
```diff
@@ -69,6 +69,9 @@ config :sentry,
   included_environments: ~w(production staging),
   environment_name: System.get_env("ENV") || "development"
 
+config :geo_postgis,
+  json_library: Poison
```
rhnonose: Should be Jason if possible.

PR #570, opened 2019-05-22
Setup price suggestion v2 client

rhnonose:
Setup the v2 for price suggestions, a client to call `priceteller`


apps/re_integrations/lib/priceteller/output.ex
```diff
@@ -0,0 +1,34 @@
+defmodule ReIntegrations.PriceTeller.Output do
+  @moduledoc """
+  Module for grouping credipronto simulator query input params
```
pmargreff: ops

PR #571, opened 2019-05-23
Add assertion to tests that might be unordered

rhnonose:
This should fix flip tests and make it easier to assert lists that we don't care about the order.

### DIFF
--- a/apps/re_web/test/support/conn_case.ex
+++ b/apps/re_web/test/support/conn_case.ex
@@ -40,13 +40,6 @@ defmodule ReWeb.ConnCase do
 
         put_req_header(conn, "authorization", "Token #{jwt}")
       end
-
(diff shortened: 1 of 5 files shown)


apps/re_web/test/support/conn_case.ex
```diff
@@ -40,6 +40,13 @@ defmodule ReWeb.ConnCase do
 
         put_req_header(conn, "authorization", "Token #{jwt}")
       end
+
+      defmacro assert_mapper_match(left, right, key_func) do
```
pmargreff: Is this definition duplicated? I guess the one living on model_case does the job.

pmargreff: You can create a new module called `CustomAsserts` and when we need to use those we can import this module and also aggregate new asserts if they're required.

PR #574, opened 2019-05-28
Drop unused statistics

rhnonose:
Since we're now gathering most stats through amplitude, it's no longer necessary to store in our database.

TODO: The `in_person_visit` stat will actually be missing since we used to sync through pipedrive's integration and will have to rebuild.


rhnonose:
Yeah, wasn't sure about how to properly replace it.
Wasn't expecting to merge this PR right away, I'll still do some tests and ask around what's the impact on clients.

PR #577, opened 2019-05-29
Add budget form mutation

rhnonose:
Add the mutation for the new budget form, without the underlying functionality so @gabiseabra can build on top of it.


apps/re_web/lib/graphql/resolvers/buyer_leads.ex
```diff
@@ -0,0 +1,8 @@
+defmodule ReWeb.Resolvers.BuyerLeads do
+  @moduledoc """
+  Resolver module for seller leads
```
pmargreff:
```suggestion
  Resolver module for buyer leads
```

PR #578, opened 2019-05-29
fix(re-integrations): fix test transactions

pmargreff:
Add ModelCase on ReIntegration allowing to run each test in an independent transaction. Also, add both repositories on Web Conn helpers. 

I totally forgot it, 'cause I used to reset the database (testing prefixes/schemas) it didn't show up `ReIntegration.Repo` entities not being cleaned up.

### DIFF
--- a/apps/re_integrations/test/support/model_case.ex
+++ b/apps/re_integrations/test/support/model_case.ex
@@ -1,5 +1,8 @@
 defmodule ReIntegrations.ModelCase do
+  @moduledoc false
+
   use ExUnit.CaseTemplate
+  alias Ecto.Adapters.SQL.Sandbox
 
   using do
     quote do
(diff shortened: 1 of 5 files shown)


apps/re_integrations/test/support/model_case.ex
```diff
@@ -0,0 +1,31 @@
...
+
+      import ReIntegrations.ModelCase
+    end
+  end
+
+  setup tags do
+    :ok = Sandbox.checkout(ReIntegrations.Repo)
+
+    unless tags[:async] do
+      Sandbox.mode(ReIntegrations.Repo, {:shared, self()})
```
rhnonose: `conn_case` is setting sandbox mode for both repos but here it isn't.

pmargreff: But this is the intention, `ReIntegrations.ModelCase` has control only over `ReIntegrations` cases, while ConnCase can touch on both connections running integration tests ...

rhnonose: Not necessarily, there might be some test case that touches both.
It's better to be safe than sorry, this kind of error is actually hard to notice when it breaks.

PR #580, opened 2019-05-30
Orulo/parse buildings to development

pmargreff:
Continuing the work started on PR #573. 

Here I'm getting a payload previously fetched from the external API, parsing the params to compatible address' and development's params and create new entities in our systems. Some highlights from this process are: 
- Making all in a transaction using [run](https://hexdocs.pm/ecto/Ecto.Multi.html#run/3). 
- Static factory setup for Building. 
---

One thing that caught my attention while writing this PR, maybe a better name to `Building` module (consequently also the table) is `BuidingPayload`, it's a better representation about what it represents, @rhnonose what are your thoughts about it?  

---

Docs don't make explicit how powerful run is, found it in some [random](https://geoffreylessel.com/2017/using-ecto-multi-to-group-database-operations/) [articles](http://engineering.teacherspayteachers.com/2018/02/22/database-transactions-or-how-i-learned-to-stop-worrying-and-love-ecto-multi.html). Using this interface we can use transactions without rewrite all our context crud operations, in the long term I guess it could be a good solution, we still writing atomic operations, and group them in multi transactions as needed, so we don't need to keep an `insert/?`, `insert_multi/?`, `update/?`, `update_multi/?` ... in each context.

### DIFF
--- a/apps/re_integrations/lib/orulo/job_queue.ex
+++ b/apps/re_integrations/lib/orulo/job_queue.ex
@@ -23,4 +23,8 @@ defmodule ReIntegrations.Orulo.JobQueue do
       error -> Logger.error(error)
     end
   end
+
+  def perform(%Multi{} = _multi, %{"type" => "parse_building_into_development", "uuid" => uuid}) do
+    Orulo.insert_development_from_building(uuid)
+  end
 end
--- a/apps/re_integrations/lib/orulo/job_queue.ex
+++ b/apps/re_integrations/lib/orulo/job_queue.ex
@@ -23,7 +23,4 @@ defmodule ReIntegrations.Orulo.JobQueue do
       error -> Logger.error(error)
     end
   end
-
-  def perform(%Multi{} = multi, %{"type" => "parse_building_into_development", "uuid" => uuid}) do
-  end
 end
--- a/apps/re_integrations/lib/orulo/job_queue.ex
+++ b/apps/re_integrations/lib/orulo/job_queue.ex
@@ -9,8 +9,7 @@ defmodule ReIntegrations.Orulo.JobQueue do
 
   alias ReIntegrations.{
     Orulo,
-    Orulo.Client,
-    Repo
+    Orulo.Client
   }
 
   alias Ecto.Multi
@@ -26,7 +25,6 @@ defmodule ReIntegrations.Orulo.JobQueue do
   end
 
   def perform(%Multi{} = multi, %{"type" => "parse_building_into_development", "uuid" => uuid}) do
-    Orulo.insert_development_from_building(uuid)
-    Repo.transaction(multi)
+    Orulo.insert_development_from_building(multi, uuid)
   end
 end
--- a/apps/re_integrations/lib/orulo/job_queue.ex
+++ b/apps/re_integrations/lib/orulo/job_queue.ex
@@ -23,4 +23,7 @@ defmodule ReIntegrations.Orulo.JobQueue do
       error -> Logger.error(error)
     end
   end
+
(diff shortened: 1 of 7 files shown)


apps/re_integrations/lib/orulo/job_queue.ex
```diff
@@ -23,4 +23,8 @@ defmodule ReIntegrations.Orulo.JobQueue do
       error -> Logger.error(error)
     end
   end
+
+  def perform(%Multi{} = _multi, %{"type" => "parse_building_into_development", "uuid" => uuid}) do
+    Orulo.insert_development_from_building(uuid)
```
rhnonose: `multi` here comes with the operation that will mark the job as done (or, delete the record), so you have to run the operation with the multi. Otherwise, it'll run multiple times.

apps/re_integrations/lib/orulo/orulo.ex
```diff
@@ -17,13 +21,29 @@ defmodule ReIntegrations.Orulo do
...
+      "type" => "parse_building_into_development",
+      "uuid" => uuid
+    })
     |> Repo.transaction()
   end
 
-  defp insert_building_on_multi(changeset, multi) do
-    Multi.insert(multi, :building, changeset)
+  def insert_development_from_building(uuid) do
+    with building <- Repo.get(Building, uuid),
```
pmargreff: @rhnonose All this block doesn't be a multi-transaction is a choice once we cannot have address identifier before transaction...

apps/re_integrations/lib/orulo/mapper.ex
```diff
@@ -0,0 +1,80 @@
...
+
+  defp convert_development_attribute({"status", status}, acc) do
+    phase = Map.get(@phase_map, status)
+    Map.put(acc, :phase, phase)
+  end
+
+  defp convert_development_attribute(_, acc), do: acc
+
+  def building_payload_into_address_params(%Building{} = %{payload: %{"address" => address}}) do
+    Enum.reduce(address, %{}, &convert_address_attribute(&1, &2))
```
rhnonose: You should filter the attributes that will be run with `Enum.take`, otherwise there will be always unnecessary loops.

apps/re_integrations/lib/orulo/mapper.ex
```diff
@@ -0,0 +1,80 @@
+defmodule ReIntegrations.Orulo.Mapper do
+  @moduledoc """
+  Module to map external structures into persistable internal structures.
+  """
+  alias ReIntegrations.{
+    Orulo.Building
+  }
+
+  def building_payload_into_development_params(%Building{} = %{payload: payload}) do
+    Enum.reduce(payload, %{}, &convert_development_attribute(&1, &2))
```
rhnonose: You should filter the attributes that will be run with `Enum.take`, otherwise there will be always unnecessary loops.

apps/re_integrations/lib/orulo/job_queue.ex
rhnonose: It's safer if you commit in the same transaction.
If `Repo.transaction(multi)` fails for some reason, `Orulo.insert_development_from_building(uuid)` will be rerun.

apps/re_integrations/lib/orulo/orulo.ex
rhnonose: Hm, might need a new job for it then.

pmargreff: Cancel my comment, made it happen with `run` now.

rhnonose:
Makes sense since we don't have a `building` entity, would be more explicit.

`Multi.run/2` is nice for some cases, yeah. I actually knew about it, forgot somehow. Good finding.

PR #583, opened 2019-06-05
Allow interest insertion without listing id

rhnonose:
Remove the need to provite `listingId` when creating an `interest`.


rhnonose:
Closed in favor of a new mutation/schema.

PR #587, opened 2019-06-06
Orulo/clone images

pmargreff:
Continuing the efforts to integrate with Orulo here we go. 

Sorry about the PR's size, took to much time to notice I was refactoring while adding new functionalities if it is really harsh to get I can try to cherry pick it. 

---

Refactor: Created a new `PayloadProcessor` module to keep all logic to process payloads once we already have them, on the initial version it uses to go on `Orulo` entry module. 

---

Pre-requisite: 
Add `orulo_id` on developments to have the reference when the development is imported. 

---

Feature: 
Using the sync flow I'm doing both read steps related to the image: 

Orulo sync diagram (1)

We get all images payloads and save them to the registry, and in another job we create images. 

The step to create images make two processes, first, upload them to our image provider and after we save the uploaded images. This part is a bit tricky and can have multiples points of failures. By now if one or multiple uploads fail, I'm logging it and continuing the process, the sync only will be considered an error if all image uploads failed. 

The problem is, it could duplicate the upload multiple times on retry. One option is to implement the delete to images on the case we faile to save images, but I really don't know how frequently it can happen on the beginning. 

I've some minor refactors and places to test, but in the big picture, all the things are here.


apps/re_integrations/lib/orulo/job_queue.ex
```diff
@@ -9,22 +9,42 @@ defmodule ReIntegrations.Orulo.JobQueue do
...
 
   def perform(%Multi{} = multi, %{"type" => "import_development_from_orulo", "external_id" => id}) do
     with {:ok, %{body: body}} <- Client.get_building(id),
          {:ok, payload} <- Jason.decode(body),
-         {:ok, _} <- Orulo.multi_building_insert(multi, %{external_id: id, payload: payload}) do
+         {:ok, _} <-
+           Orulo.multi_building_payload_insert(multi, %{external_id: id, payload: payload}) do
+    else
+      {:error, error} -> Logger.error(error)
+      error -> Logger.error(error)
```
rhnonose: Inside a `ecto_job`, if you want to interrupt a job, you have to raise an error.
If you don't, the job will stay `IN_PROGRESS`, expire and will try indefinitely.

apps/re/lib/developments/development.ex
```diff
@@ -31,7 +32,7 @@ defmodule Re.Development do
 
   @required ~w(name phase builder description address_id)a
 
-  @optional ~w(floor_count units_per_floor elevators)a
+  @optional ~w(floor_count units_per_floor elevators orulo_id)a
```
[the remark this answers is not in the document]
pmargreff: Yes, we can, currently, we already can add development's listings, from `/developments` on garagem, and it's independent of orulo.

apps/re_integrations/lib/orulo/payloads_processor.ex
```diff
@@ -0,0 +1,113 @@
...
+      {:ok, transaction}
+    else
+      err -> err
+    end
+  end
+
+  defp insert_transaction(
+         multi,
+         address_params,
+         development_params = %{orulo_id: orulo_id}
```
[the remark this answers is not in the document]
pmargreff: good job bot

apps/re_integrations/lib/orulo/job_queue.ex
pmargreff: I guess it's ok to retry on this case, the database can be out or something like this. The only thing doesn't look good is to retry forever, I'll define the max number to retry on these cases.

rhnonose: There's a max number of retries as a default, you shouldn't worry about that.
The thing is, it'll keep retrying without raising an error. Maybe you should check `ecto_job`s docs for a more detailed view, but this will leave the job record pending AFAIK.

apps/re_integrations/lib/orulo/payloads_processor.ex
```diff
@@ -66,6 +66,10 @@ defmodule ReIntegrations.Orulo.PayloadsProcessor do
     |> Multi.run(:insert_images, fn _repo, _changes ->
       case images_upload_response do
         [] ->
+          Sentry.capture_message("images_upload_failed",
```
pmargreff: @rhnonose we can send other messages than exceptions to sentry with [Sentry.capture_message](https://hexdocs.pm/sentry/readme.html#capture-non-exception-events).

PR #588, opened 2019-06-06
Import orulo tags

rhnonose:
Core:
- Import tags from orulo's building payload
  - Enqueue a job for doing that asynchronously.
  - Inside the attribute `features`, there's a list of strings containing development features
  - Use a hard-coded map (`TagMapper`) to map their tags to ours. Not mapped tags will be ignored
- Not handled: if there's a `building_payload` for the same development, tags will be inserted again and break a unique constraint. If there's a case where that happens, we should `delete_all` tags from that development prior to inserting or handling conflicts (not sure if `Multi.insert_all` would do that properly, maybe would have to be through individual inserts

TODO:
  - Expose development tags through the API

Extra:
- Refactor `PayloadsProcessor`.insert_development_from_building_payload/2` for readability

### DIFF
--- a/apps/re_integrations/lib/orulo/payloads_processor.ex
+++ b/apps/re_integrations/lib/orulo/payloads_processor.ex
@@ -17,39 +17,37 @@ defmodule ReIntegrations.Orulo.PayloadsProcessor do
   }
 
   def insert_development_from_building_payload(multi, building_uuid) do
-    with building <- Repo.get(BuildingPayload, building_uuid),
-         address_params <- Mapper.building_payload_into_address_params(building),
-         development_params <- Mapper.building_payload_into_development_params(building),
-         {:ok, transaction} <-
-           insert_transaction(multi, address_params, development_params) do
-      {:ok, transaction}
-    else
-      err -> err
-    end
-  end
+    building = Repo.get(BuildingPayload, building_uuid)
 
-  defp insert_transaction(
-         multi,
-         address_params,
-         %{orulo_id: orulo_id} = development_params
-       ) do
     multi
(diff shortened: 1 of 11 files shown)


apps/re_integrations/lib/orulo/payloads_processor.ex
```diff
@@ -115,4 +126,27 @@ defmodule ReIntegrations.Orulo.PayloadsProcessor do
         error
     end
   end
+
+  def process_orulo_tags(multi, uuid) do
```
pmargreff: While doing images I made a `PayloadProcessor` module to do this kind of job, you could move this function to keep these transformations togheter.

apps/re_integrations/lib/orulo/tag_mapper.ex
```diff
@@ -0,0 +1,46 @@
...
+  Module to map orulo's tags into our tags.
+  """
+  alias ReIntegrations.{
+    Orulo.BuildingPayload
+  }
+
+  def map_tags(%BuildingPayload{} = %{payload: %{"features" => features}}) do
+    features
+    |> Enum.reduce([], &convert_tag(&1, &2))
+    |> Enum.dedup()
```
pmargreff: TIL

apps/re_integrations/lib/orulo/payloads_processor.ex
rhnonose: Hmm, I think it should be in the same module but renamed to `BuildingPayload.Processor` or something, because features is inside the same payload.

pmargreff: Ignore my comment, should be sleepy while reviewing the code.

PR #589, opened 2019-06-07
Add codecov

pmargreff:
Once it uses coveralls same settings we need no config changes to make it work. :tada: 	

As it hasn't run in master yet and it's the default branch it has no status, but you can navigate in the [current branch](https://codecov.io/gh/emcasa/backend/branch/add-codecov). The interface looks pretty good, expect it help us with better visibility about coverage.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,6 +52,8 @@ jobs:
           name: Wait for DB
           command: dockerize -wait tcp://localhost:5432 -timeout 1m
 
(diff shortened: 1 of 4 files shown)


pmargreff:
Also, review comments and checks will be included once we have a master, and [their docs are really good](https://docs.codecov.io/docs/codecov-delta).

rhnonose:
🎉 ayyy thanks

PR #593, opened 2019-06-07
Get facebook retailer item

rhnonose:
Since facebook forms doesn't send `retailer_item_id`, we need to fetch the lead to retrieve it.

Here I introduced `mockery` to be able to inject dynamically `listing_id` as `retailer_item_id` to be returned from the mocked HTTP call

### DIFF
--- a/apps/re/lib/buyer_leads/facebook.ex
+++ b/apps/re/lib/buyer_leads/facebook.ex
@@ -48,6 +48,8 @@ defmodule Re.BuyerLeads.Facebook do
   def buyer_lead_changeset(nil), do: raise("Leads.FacebookBuyer not found")
 
   def buyer_lead_changeset(lead) do
+    {:ok, listing_uuid} = extract_listing_uuid(lead.lead_id)
+
     BuyerLead.changeset(%BuyerLead{}, %{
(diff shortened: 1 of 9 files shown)


apps/re/lib/buyer_leads/facebook.ex
```diff
@@ -72,4 +76,19 @@ defmodule Re.BuyerLeads.Facebook do
...
   end
+
+  defp extract_listing_uuid(lead_id) do
+    with {:facebook_client_call, {:ok, %{body: body}}} <-
+           {:facebook_client_call, FacebookClient.get_lead(lead_id)},
+         {:ok, %{"retailer_item_id" => listing_id}} <- Jason.decode(body),
+         {:ok, listing} <- Re.Listings.get(listing_id) do
+      {:ok, listing.uuid}
+    else
+      {:facebook_client_call, error} ->
```
pmargreff: Shouldn't we process the lead without the listing reference on these cases instead of losing them? You can use the same approach we discussed earlier and launch a message to sentry...

rhnonose: This is for an HTTP call error, so I figured we should retry it.

PR #594, opened 2019-06-08
Orulo/reject duplicated building payloads

pmargreff:
Reject client request to sync new building if building payload already exists. So we don't reintroduce one building which was already inserted. 

This is partial, I need to solve the problem for the scenario: if two different clients send the request at the same time, so both will be accepted and one of failing on save development step (cause we have a constraint on developments). 

I see two different approaches here: 
- Add a unique constraint for `external_id` in `building_payloads` table, I'm no confident about this one cause we could keep the payloads to updates on this table as well (not done yet). 
- Or to check if a building payload already exists [on this multi](https://github.com/emcasa/backend/pull/594/files#diff-794db1e40f3a626f1d3f240c77626081R38), the problem is this one can also get a race condition once two different transactions can return false to `building_payload_exists?/1` at the same time, the timeframe here need to be tight to this situation happen, but still a possible issue. 

What do you think about this possible problem, should we solve this right now or expect it to be a "real problem" to deal with?

### DIFF
--- a/apps/re_integrations/lib/orulo/orulo.ex
+++ b/apps/re_integrations/lib/orulo/orulo.ex
@@ -2,6 +2,9 @@ defmodule ReIntegrations.Orulo do
   @moduledoc """
   Context module to use importers.
   """
+
+  import Ecto.Query, only: [from: 2]
+
   alias ReIntegrations.{
     Orulo.BuildingPayload,
     Orulo.ImagePayload,
@@ -50,5 +53,10 @@ defmodule ReIntegrations.Orulo do
       "uuid" => uuid
     })
     |> Repo.transaction()
+  end
+
+  def building_payload_synced?(external_id) do
+    from(bp in BuildingPayload, where: bp.external_id == ^external_id)
+    |> Repo.exists?()
   end
 end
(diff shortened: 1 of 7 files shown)


rhnonose:
This modification plus a unique constraint on `external_id` should do the trick.
I just didn't understand from your explanation how the API works.
If they have an update on `building`'s attributes, do they send as a new building or an update call? (which would be the standard/sane approach, but seems like it's not quite that way).
(assuming) if the update is done by sending a new building with the same ID like a new building, then we'll have to have some structure to resolve the race condition.

pmargreff:
> ... an update call? (which would be the standard/sane approach, but seems like it's not quite that way).

Do you mean a webhook? If yes, we've talked about it as a future possibility, but we haven't it today. 

They update the same building and don't send a clue about this action. A strategy to detect update is to run a daily/weekly job and check if the `updated_at` from the payload is more recent than our `created_at`, and if it's we get the payload and trigger a pipeline to update the data on our side (same strategy to units), otherwise we discard the payload.

rhnonose:
All right then.
We should have both checks and be attentive to errors.

PR #595, opened 2019-06-10
Filters/add is release

pmargreff:
Add filter to new/used listings. Frontend use to infer this rule from `developments` associations, it works but could cause some bad dependencies in the future. 

We already set all releases from developments with `is_release: true` but old ones don't use to be set. I'm setting the default for `is_release` as false now to use this attribute as a filter to differentiate primary and secondary market.  

Also, this PR made me see how confuse our test on listings query with filters is, I'll take some time to refactor trying to make it simple, and transfer the logic tested there to `filters_test` ...

### DIFF
--- a/apps/re/priv/repo/migrations/20190610140649_update_is_release_on_listings.exs
+++ b/apps/re/priv/repo/migrations/20190610140649_update_is_release_on_listings.exs
@@ -4,15 +4,8 @@ defmodule Re.Repo.Migrations.UpdateIsReleaseOnListings do
   require Ecto.Query
 
   def up do
-    listing_ids =
-      "listings"
-      |> Ecto.Query.from([l], select: l.id, where: is_nil(l.is_release))
-      |> Re.Repo.all()
-
     "listings"
-    |> Ecto.Query.from([l],
-      where: l.id in ^listing_ids
-    )
+    |> Ecto.Query.where([l], is_nil(l.is_release))
     |> Re.Repo.update_all(set: [is_release: false])
   end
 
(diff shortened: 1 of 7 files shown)


apps/re/priv/repo/migrations/20190610140649_update_is_release_on_listings.exs
```diff
@@ -0,0 +1,19 @@
...
+
+  def up do
+    listing_ids =
+      Ecto.Query.from(l in "listings", select: l.id, where: is_nil(l.is_release))
+      |> Re.Repo.all()
+
+    Ecto.Query.from(i in "listings",
+      where: i.id in ^listing_ids
+    )
+    |> Re.Repo.update_all(set: [is_release: false])
```
rhnonose: I think you can put these two queries together:
```
"listings"
|> Ecto.Query.where([l], is_nil(l.is_release))
|> Re.Repo.update_all(set: [is_release: false])
```

apps/re/test/listings/listings_test.exs
```diff
@@ -427,6 +427,24 @@ defmodule Re.ListingsTest do
...
+               Listings.insert(@insert_listing_params,
+                 address: address,
+                 user: user,
+                 owner_contact: owner_contact
+               )
+
+      assert retrieved_listing = Repo.get(Listing, inserted_listing.id)
+      assert retrieved_listing.uuid
+      assert retrieved_listing.is_release == false
+      assert retrieved_listing.is_exportable == true
```
rhnonose: When asserting booleans, just `assert boolean` or `refute boolean` is enough (and more readable)

pmargreff: I disagree, they could easily return false positive. 

If we're using `refute x.bla` it will return success to `nil` and it's something we don't want here once we're checking if it has value...

rhnonose: It's a boolean field with a default value, so that's hardly a problem.
But in any case, better safe than sorry.

PR #597, opened 2019-06-10
Refactor filters tests

pmargreff:
Refactor filter test to not keep going to a painful path. 

- Remove redundant assert on count (they're covered by `assert_mapper_match`). 
- Remove logic assertions from web tests and keep them on filter tests in a more unitary way. 
- Add o new test to apply all filters at once check if the tests are being cumulative.
- Keep the simplest case on web tests with all filters to ensure the params are being filtered.


rhnonose:
Love this ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️

PR #601, opened 2019-06-11
Add vila mariana

rhnonose:
- Add a task to insert new district
- Add some normalization for district aliases
- Add some tests for the task, mainly to use as a future reference for tasks since we might be using more frequently in favor of migrations

### DIFF
--- a/apps/re/lib/addresses/district.ex
+++ b/apps/re/lib/addresses/district.ex
@@ -21,8 +21,8 @@ defmodule Re.Addresses.District do
     timestamps()
   end
 
-  @required ~w(state city name description)a
-  @optional ~w(status)a
+  @required ~w(state city name)a
+  @optional ~w(status description)a
   @params @required ++ @optional
 
   @sluggified_attr ~w(state city name)a
--- a/apps/re/lib/addresses/district.ex
+++ b/apps/re/lib/addresses/district.ex
@@ -15,7 +15,7 @@ defmodule Re.Addresses.District do
     field :state_slug, :string
     field :city_slug, :string
     field :name_slug, :string
-    field :description, :string
+    field :description, :string, default: ""
     field :status, :string, default: "inactive"
 
     timestamps()
--- a/apps/re/lib/addresses/neighborhoods.ex
+++ b/apps/re/lib/addresses/neighborhoods.ex
@@ -76,7 +76,7 @@ defmodule Re.Addresses.Neighborhoods do
   def is_covered(neighborhood) do
     neighborhood
     |> sluggify_attributes()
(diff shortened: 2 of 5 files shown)


apps/re/lib/addresses/neighborhoods.ex
```diff
@@ -89,9 +89,43 @@ defmodule Re.Addresses.Neighborhoods do
 
   defp normalize(
          %{city_slug: "sao-paulo", state_slug: "sp", neighborhood_slug: "pompeia"} = neighborhood
-       ) do
-    %{neighborhood | neighborhood_slug: "vila-pompeia", neighborhood: "Vila Pompeia"}
-  end
+       ),
+       do: %{neighborhood | neighborhood_slug: "vila-pompeia", neighborhood: "Vila Pompeia"}
+
+  defp normalize(
```
pmargreff: A suggestion here, in these cases`remap_neighborhood`, `reroute_neighborhood`, `rematch_neighborhood` or something on this direction sounds meaningful than `normalize`

apps/re/lib/mix/tasks/addresses/add_district.ex
```diff
@@ -0,0 +1,33 @@
...
+        state: "SP",
+        city: "São Paulo",
+        status: "active"
+      })
+      |> Re.Repo.insert()
+
+    Logger.info("Inserted district named #{district.name}")
+  end
+
+  def insert(params) do
```
pmargreff: Is this fn right?

rhnonose: No, copy-paste problems 😅

rhnonose:
The districts are inserted as is, this is just to check if it's covered or not.
We definitely should check the "browseability" of those aliases for ssure.

apps/re/lib/addresses/district.ex
```diff
@@ -21,8 +21,8 @@ defmodule Re.Addresses.District do
     timestamps()
   end
 
-  @required ~w(state city name description)a
-  @optional ~w(status)a
+  @required ~w(state city name)a
+  @optional ~w(status description)a
```
pmargreff: Couldn't this cause a regression on the client?

rhnonose: What kind of regression?

pmargreff: An `undefined` on neighborhood description.

rhnonose: Our frontend removed recently, but we might send a default empty string just in case, I'll check with them.

PR #602, opened 2019-06-12
sync typologies

pmargreff:
Get payloads for a building's typology.

Some late fixes: 
- Alter the id from `integer` to `string` and change the name from `external_id` to `building_id` to be more specific about what object we're referring (I'll replicate the changes for the other payloads before start to use in prod).  

Other: 
- Add a helper to test specific jobs enqueued, don't know if it's the best approach, but help to false positives while testing jobs.

### DIFF
--- a/apps/re_integrations/lib/orulo/job_queue.ex
+++ b/apps/re_integrations/lib/orulo/job_queue.ex
@@ -37,6 +37,17 @@ defmodule ReIntegrations.Orulo.JobQueue do
     end
   end
 
+  def perform(%Multi{} = multi, %{"type" => "fetch_typology", "external_id" => id}) do
+    with {:ok, %{body: body}} <- Client.get_typologies(id),
+         {:ok, payload} <- Jason.decode(body),
+         {:ok, _} <-
+           Orulo.insert_typologies_payload(multi, %{external_id: id, payload: payload}) do
+    else
+      {:error, error} -> Logger.error(error)
+      error -> Logger.error(error)
+    end
+  end
+
   def perform(%Multi{} = multi, %{"type" => "parse_building_into_development", "uuid" => uuid}) do
     PayloadProcessor.insert_development_from_building_payload(multi, uuid)
   end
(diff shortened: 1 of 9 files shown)


apps/re_integrations/lib/orulo/job_queue.ex
```diff
@@ -37,6 +37,20 @@ defmodule ReIntegrations.Orulo.JobQueue do
...
   end
 
+  def perform(%Multi{} = multi, %{"type" => "fetch_typology", "building_id" => id}) do
+    with {:ok, %{body: body}} <- Client.get_typologies(id),
+         {:ok, payload} <- Jason.decode(body),
+         {:ok, _} <-
+           Orulo.insert_typologies_payload(multi, %{
+             building_id: Integer.to_string(id),
+             payload: payload
+           }) do
```
rhnonose: Instead of leaving `do` empty, you could put the final clause there.

apps/re_integrations/lib/orulo/job_queue.ex
```diff
@@ -37,6 +37,20 @@ defmodule ReIntegrations.Orulo.JobQueue do
...
+    else
+      {:error, error} -> Logger.error(error)
+      error -> Logger.error(error)
```
rhnonose: You could put both of theses in the same line with `Kernel.inspect` since we don't need the return anyway.

pmargreff:
> I'm starting to think it might be better if we put all payloads in the same building table since it all pertains to the same building instead of having to fetch independently, not sure if you got that feeling.

Absolutely, I start to think the same. After implementing the parser to transform topologies on listings I can refactor to use the same table to payloads and see how it looks like.

apps/re_integrations/lib/orulo/job_queue.ex
pmargreff: I'm replicating the last one, keep it on with so we can show errors in a standard way .

PR #603, opened 2019-06-12
Miscellaneous improvements

rhnonose:
- Update `ecto_job` to use `requeue` function
- Remove old tag task and put in seeds
- Add `on_delete: :delete_all` to enable deleting users
- Deprecate some unused inputs
- Send errors to sentry when jobs fail


apps/re/lib/developments/development.ex
```diff
@@ -43,9 +43,7 @@ defmodule Re.Development do
     struct
     |> cast(params, @required ++ @optional)
     |> validate_required(@required)
-    |> validate_inclusion(:phase, @phases,
-      message: "should be one of: [#{Enum.join(@phases, " ")}]"
-    )
+    |> validate_inclusion(:phase, @phases, message: "invalid value")
```
pmargreff: I really think the original is useful, mainly to help clients to find errors without backend intervention. 

Another option to centralize logic on `ChangesetHelper`, creating a `validate_inclusion_with_valid_values_on_message` function. 

And yet if you want to remove I guess the original message (`is invalid`) is good enough and have the same level of communication.

pmargreff:
Also, didn't find anything related to `requeue` on `ecto_job`, did you committed it?

rhnonose:
@pmargreff it's for triggering manually for now.
I used to just `Repo.update_all` the records (which is kinda what tha function does), but I took the opportunity to update the dependencies.

apps/re/lib/developments/development.ex
rhnonose: This `should be one of...` message was due to ecto not returning the possible values that `validate_inclusion` was checking.
Now that it does, it's not necessary anymore.
But yeah, I forgot the original was `is invalid`, I'll change.

PR #604, opened 2019-06-12
Fix facebook access config

rhnonose:
Facebook access token for fetching lead was in the wrong place (under `Re.Repo` config 🤦‍♂ )

Extra:
  Add a utility function, `requeue_all` for jobs


apps/re/lib/buyer_leads/job_queue.ex
```diff
@@ -82,6 +82,16 @@ defmodule Re.BuyerLeads.JobQueue do
 
   def perform(_multi, job), do: raise("Job type not handled. Job: #{Kernel.inspect(job)}")
 
+  def requeue_all(multi) do
```
pmargreff: Not now, but in the future, this could be extracted to a queue helper module.

rhnonose: I may open a PR to the lib itself.

PR #606, opened 2019-06-13
Add user url to buyer leads

rhnonose:
~The buyer leads processing was a little messy with a bunch of functions pattern matching possibly new values, so I grouped into `location` and `user`, reducing a little the branching conditions.
This was prompted by the necessity to add the user URL to buyer lead (incoming PR).~

Since the change itself was minuscule because of the refactoring, here it is altogether.

### DIFF
--- a/apps/re/lib/buyer_leads/facebook.ex
+++ b/apps/re/lib/buyer_leads/facebook.ex
@@ -54,7 +54,6 @@ defmodule Re.BuyerLeads.Facebook do
       %{
         name: lead.full_name,
         email: lead.email,
-        phone_number: lead.phone_number,
         origin: "facebook",
         budget: lead.budget,
         neighborhood: lead.neighborhoods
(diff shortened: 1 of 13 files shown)


apps/re/lib/buyer_leads/facebook.ex
```diff
@@ -49,46 +50,68 @@ defmodule Re.BuyerLeads.Facebook do
...
+      |> put_location(lead)
+      |> put_user_info(lead)
+
+    BuyerLead.changeset(%BuyerLead{}, params)
+  end
+
+  defp put_location(params, %{lead_id: lead_id, location: location}) do
+    lead_id
+    |> get_listing()
+    |> case do
```
pmargreff: this `case` can be extracted to a function called `put_location`/`put_address`

apps/re/lib/buyer_leads/facebook.ex
```diff
@@ -49,46 +50,68 @@ defmodule Re.BuyerLeads.Facebook do
...
+        Map.put(params, :location, get_location(location))
+    end
+  end
+
+  defp put_user_info(params, %{phone_number: nil}), do: params
+
+  defp put_user_info(params, %{phone_number: phone_number}) do
+    phone_number
+    |> Users.get_by_phone()
+    |> case do
```
pmargreff: this case can be extracted to a `put_user_info` function.

apps/re/lib/buyer_leads/facebook.ex
```diff
@@ -49,46 +50,68 @@ defmodule Re.BuyerLeads.Facebook do
...
-         {:ok, listing} <- Re.Listings.get(listing_id) do
-      {:ok, listing.uuid}
-    else
-      error ->
-        Sentry.capture_message("error when processing facebook buyer lead",
-          extra: %{lead_id: lead_id, error: error}
-        )
-
-        {:ok, nil}
+  defp get_retailer_item_id(body) do
```
pmargreff: `get_listing_id`, `get_listing_id_from_body` or something like this is more meaningful once it tells you about what you're picking instead where it's.

apps/re/lib/buyer_leads/grupozap.ex
```diff
@@ -48,42 +48,51 @@ defmodule Re.BuyerLeads.Grupozap do
...
+      |> put_user_info(gzb)
+
+    BuyerLead.changeset(%BuyerLead{}, params)
   end
 
-  defp get_location({:ok, %{address: address}}), do: "#{address.city_slug}|#{address.state_slug}"
-  defp get_location(_), do: "unknown"
-
-  defp concat_phone_number(%{ddd: _ddd, phone: nil}), do: "not informed"
+  defp put_location(params, gzb) do
```
pmargreff: `gzb` isn't a good name once it depends on some predisposition to look around to get the meaning. `lead` is good enough, `grupozap_buyer` and `grupozap_buyer_lead` help to get the context without any search. 
Didn't call it on the first PR cause it was only in the changeset, now it is spreading around.

apps/re/lib/interests/interest.ex
```diff
@@ -46,35 +46,41 @@ defmodule Re.Interest do
...
+  defp put_user_info(params, %{phone: nil}), do: params
 
-  defp format(phone_number), do: String.replace(phone_number, ["(", ")", "-", " "], "")
+  defp put_user_info(params, interest) do
+    phone_number = String.replace(interest.phone, ["(", ")", "-", " "], "")
 
-  defp extract_user_uuid(nil), do: nil
+    phone_number
+    |> Users.get_by_phone()
+    |> case do
```
pmargreff: same call from above to extract this case to `get_info`.

PR #607, opened 2019-06-14
Persist price suggestion

rhnonose:
Add an `ecto_job` for fetching and persisting suggested price when:
  - Listing is activated
  - Listing is updated

Also fetch and persist when listing is queried but `suggested_price` is `nil`

Listing insertion not handled since, in our current use cases, a listing is added barely empty at first.

This is sort of a middle-of-the-road workaround to avoid doing rate-limited/batch requests to priceteller since it's a lot of work for a one-time use.

### DIFF
--- a/apps/re/lib/listings/job_queue.ex
+++ b/apps/re/lib/listings/job_queue.ex
@@ -28,12 +28,12 @@ defmodule Re.Listings.JobQueue do
       error ->
         error
     end
-    |> log_error()
+    |> handle_error()
   end
 
-  defp log_error({:ok, result}), do: {:ok, result}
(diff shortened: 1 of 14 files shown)


apps/re/lib/listings/job_queue.ex
```diff
@@ -0,0 +1,43 @@
...
+
+      error ->
+        error
+    end
+    |> log_error()
+  end
+
+  defp log_error({:ok, result}), do: {:ok, result}
+
+  defp log_error(error) do
```
pmargreff: `notify_error` has a better cohesion since `log` has a clear meaning in our context.

PR #608, opened 2019-06-14
Fetch typologies

pmargreff:
This is the very last pass to have all workflow running and up. 

### Fixes: 

- Keys for `tags_process` job was adjusted do create a job with `uuid` instead `external_id`.

### Some context: 

To create one unit/listing we need information which is both on typology and unit payload, so we need first get them and after merge. Typologies contain some info as rooms, bathrooms and garage spots, while one typology can contain multiple units which keep complement and price information. 

### Feature: 

This is the very first step to the end of the process: Here we need to fetch typologies and units to create our listings, by this very first moment I'm only using typology to validate the steps and simplify the implementation (only as a mid-step). 

So it's creating our units from typologies, what isn't the correct map (need to fetch and merge units info). 
In the final version, we'll fetch all units for each typology and these ones will create our `units/listings`. 

### Workflow overview:

Orulo sync diagram (3)

---
- Green points are ones who are already implemented. 
- blue is what happens on the current point of this PR.  
- red includes what should happen at the end of this PR.

### DIFF
--- a/apps/re_integrations/lib/orulo/orulo.ex
+++ b/apps/re_integrations/lib/orulo/orulo.ex
@@ -82,18 +82,19 @@ defmodule ReIntegrations.Orulo do
   def bulk_insert_unit_payload_forking_multi(%Multi{} = multi, building_id, responses) do
     unit_multies =
       Enum.map(responses, fn response ->
-        with {typology_id, {:ok, %{body: payload}}} <- response do
-          #  {:ok, payload} <- Jason.decode(body),
+        with {typology_id, {:ok, %{body: body}}} <- response,
+             {:ok, payload} <- Jason.decode(body) do
           insert_unit_payload(
             Multi.new(),
             %{
               building_id: building_id,
-              typology_id: Integer.to_string(typology_id),
+              typology_id: typology_id,
               payload: payload
             }
           )
         else
-          error -> Logger.error("Error on units request:  #{Kernel.inspect(error)}")
+          error ->
+            Logger.error("Error on units request:  #{Kernel.inspect(error)}")
         end
       end)
 
--- a/apps/re_integrations/lib/orulo/orulo.ex
+++ b/apps/re_integrations/lib/orulo/orulo.ex
@@ -4,12 +4,14 @@ defmodule ReIntegrations.Orulo do
   """
 
   import Ecto.Query, only: [where: 2]
+  require Logger
 
   alias ReIntegrations.{
     Orulo.BuildingPayload,
     Orulo.Client,
     Orulo.ImagePayload,
     Orulo.TypologyPayload,
+    Orulo.UnitPayload,
     Orulo.JobQueue,
     Repo
   }
@@ -75,6 +77,48 @@ defmodule ReIntegrations.Orulo do
       "uuid" => uuid
     })
     |> Repo.transaction()
+  end
+
+  def bulk_insert_unit_payload_forking_multi(%Multi{} = multi, responses) do
+    multies =
+      Enum.map(responses, fn response ->
+        #  {:ok, payload} <- Jason.decode(body),
+        with {typology_id, {:ok, %{body: payload}}} <- response do
+          insert_unit_payload(
+            Multi.new(),
+            %{
+              building_id: "1",
+              typology_id: Integer.to_string(typology_id),
+              payload: payload
+            }
+          )
+        else
+          error -> Logger.error("Error on units request:  #{Kernel.inspect(error)}")
+        end
+      end)
+
+    Enum.reduce(multies, multi, fn unit_multi, acc ->
+      Multi.prepend(acc, unit_multi)
+    end)
+    |> ReIntegrations.Repo.transaction()
+  end
+
+  def insert_unit_payload(multi, %{typology_id: typology_id} = params) do
+    insert_key = "insert_units_for_typology_#{typology_id}" |> String.to_atom()
+    process_key = "process_units_for_typology_#{typology_id}" |> String.to_atom()
+
+    changeset =
+      %UnitPayload{}
+      |> UnitPayload.changeset(params)
+
+    uuid = Changeset.get_field(changeset, :uuid)
+
+    multi
+    |> Multi.insert(insert_key, changeset)
+    |> JobQueue.enqueue(process_key, %{
+      "type" => "process_units",
+      "uuid" => uuid
+    })
   end
 
   def building_payload_synced?(external_id) do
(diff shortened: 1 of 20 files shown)


pmargreff:
cc @gabiseabra

apps/re_integrations/lib/orulo/payload_processor/typologies.ex
```diff
@@ -0,0 +1,47 @@
...
+
+    %{payload: %{"typologies" => typologies}} = typology_payload
+
+    multi
+    |> insert_units_from_typologies(typologies, development)
+    |> Repo.transaction()
+  end
+
+  defp insert_units_from_typologies(multi, typologies, development) do
+    Multi.run(multi, :insert_units, fn _repo, _changes ->
```
rhnonose: What do you expect here if one of the insertion fails?
Since you're not checking for insertion errors, you might have some fail silently.
If all should fail, should do multiple `Multi.insert`s.
If just one should fail, you should do multiple jobs.

pmargreff: I have no sure about what we could consider a failure but will think about it and offer some alternatives here.

rhnonose: Invalid payload and invalid data is pretty straight-forward.
Also for "valid" data. For example, you could have a unit with price less than `250_000` that would break insertion.

apps/re_integrations/lib/orulo/orulo.ex
```diff
@@ -69,16 +72,67 @@ defmodule ReIntegrations.Orulo do
...
-    |> JobQueue.enqueue(:parse_typologies, %{
-      "type" => "process_typologies",
+    |> JobQueue.enqueue(:fetch_units, %{
+      "type" => "fetch_units",
       "uuid" => uuid
     })
     |> Repo.transaction()
   end
 
+  def bulk_insert_unit_payload_forking_multi(%Multi{} = multi, responses) do
```
rhnonose: That's a loaded name, should be split 😅

apps/re_integrations/lib/orulo/payload_processor/typologies.ex
pmargreff: I'm login these cases now, in the same way, we're doing with images. Process all, info split them into successful and fail cases, log failed ones and considering the process as a success. 

When we find some failures here (probably 'cause validations), we should stop to debate if what we consider an invalid should be flexibility by our model or if it should keep the same and really ignore these cases...

pmargreff:
The PR is ready to review, I've tried to fix pointed issues made in previously. Most of the time on failures I'm sending some notifications around and continuing the process. As it could import a bunch of listings at once, I think at least in the begging have some values out of range for attributes 'cause provider has different rules/validations compared with our ones is something expected.

apps/re_integrations/priv/repo/migrations/20190614180206_create_orulo_unit_payloads.exs
```diff
@@ -0,0 +1,13 @@
+defmodule ReIntegrations.Repo.Migrations.CreateOruloUnitPayloads do
+  use Ecto.Migration
+
+  def change do
+    create table(:orulo_unit_payloads, primary_key: false) do
+      add :uuid, :uuid, primary_key: true
+      add :building_id, :string
+      add :typology_id, :string
+      add :payload, :map, null: false, default: %{}
+      timestamps()
```
rhnonose: We should insert timestamps with utc type: `timestamps(type: :timestamptz)`

apps/re_integrations/lib/orulo/unit_payload.ex
```diff
@@ -0,0 +1,30 @@
...
+
+  @schema_prefix "re_integrations"
+  @primary_key {:uuid, :binary_id, autogenerate: false}
+
+  schema "orulo_unit_payloads" do
+    field :building_id, :string
+    field :typology_id, :string
+    field :payload, :map
+
+    timestamps()
```
rhnonose: We should use utc type `timestamps(type: :utc_datetime)`

apps/re_integrations/test/orulo/payload_processor_test.exs
```diff
@@ -95,6 +82,113 @@ defmodule ReIntegrations.Orulo.PayloadProcessorTest do
...
+      assert inserted_unit_2.bathrooms == 2
+      assert inserted_unit_2.suites == 1
+      assert inserted_unit_2.garage_spots == 2
+      assert inserted_unit_2.garage_type == "unknown"
+      assert inserted_unit_2.complement == "100"
+      assert inserted_unit_1.status == "active"
+    end
+
+    @tag capture_log: true
+    test "log errors when try to add an unit with invalid attributes" do
```
rhnonose: You should add a test (or modify this one) for a successful insertion together with a failed one to see if one fails and other succeeds.

apps/re_integrations/test/orulo/typology_mapper_test.exs
```diff
@@ -0,0 +1,38 @@
...
+      params = TypologyMapper.typology_payload_into_unit_params(typology, unit)
+
+      assert params == %{
+               price: unit |> Map.get("price") |> round(),
+               area: unit |> Map.get("private_area") |> round(),
+               rooms: Map.get(typology, "bedrooms"),
+               bathrooms: Map.get(typology, "bathrooms"),
+               suites: Map.get(typology, "suites"),
+               garage_spots: Map.get(typology, "parking"),
+               complement: Map.get(unit, "reference")
```
rhnonose: Not sure why you're doing a `Map.get/2` on test data that you are hardcoding, it should be hard-coded in this assertion too.

apps/re_integrations/lib/orulo/orulo.ex
```diff
@@ -65,14 +68,78 @@ defmodule ReIntegrations.Orulo do
...
+  defp extract_response_attributes({typology_id, {:ok, %{body: body}}}) do
+    case Jason.decode(body) do
+      {:ok, payload} -> {:ok, typology_id, payload}
+      {:error, error} -> {:error, error}
+    end
+  end
+
+  defp unit_process_multi(%{typology_id: typology_id} = params) do
+    insert_key = "insert_units_for_typology_#{typology_id}" |> String.to_atom()
+    process_key = "process_units_for_typology_#{typology_id}" |> String.to_atom()
```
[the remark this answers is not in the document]
pmargreff: good catch.

apps/re_integrations/test/orulo/typology_mapper_test.exs
pmargreff: probably cause in the first version I used factories on test, I'll change it.

PR #609, opened 2019-06-14
Add `on_delete: :nilify_all` to allow deleting old data

rhnonose:


rhnonose:
I prefered the use of `nilify_all` for images because we'll retain the reference in the database to delete on the source when we do a cleanup old images on cloudinary.
Also, the downsides pointed do not apply to this case since those records don't have any other  references.
Also, it's not clear in the docs if I can alter a table through migration, only when creating a new. I'll try to change the reference, but seems unlikely.

rhnonose:
Yep, tried modifying the foreign key, didn't work.
I might upload this just to cleanup old listings that are bugging us and then remove the code to avoid problems.

PR #611, opened 2019-06-15
Add retry logic to handle priceteller timeouts

rhnonose:
Use `retry` when requesting price suggestions.
We still didn't decide how to handle when the service is not accessible, so I'll return a `Timeout` message for now.

### DIFF
--- a/apps/re/test/price_suggestions/price_suggestions_test.exs
+++ b/apps/re/test/price_suggestions/price_suggestions_test.exs
@@ -119,5 +119,26 @@ defmodule Re.PriceSuggestionsTest do
              end) =~
                ":invalid_input in priceteller"
     end
+
+    @tag capture_log: true
+    test "should handle timeout error" do
+      %{uuid: uuid} =
+        listing =
+        insert(
+          :listing,
+          rooms: 2,
+          area: 80,
+          address: build(:address, state: "MS", city: "Mah City", street: "Mah Street"),
+          garage_spots: 1,
+          bathrooms: 1,
+          suggested_price: nil
+        )
+
+      mock(HTTPoison, :post, {:error, %{reason: :timeout}})
+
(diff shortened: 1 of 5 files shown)


apps/re_integrations/lib/priceteller/price_teller.ex
```diff
@@ -2,6 +2,8 @@ defmodule ReIntegrations.PriceTeller do
   @moduledoc """
   Context module for price suggestion
   """
+  use Retry
```
pmargreff: this module shouldn't be in our dependencies?

apps/re/test/price_suggestions/price_suggestions_test.exs
```diff
@@ -119,5 +119,26 @@ defmodule Re.PriceSuggestionsTest do
...
+
+    @tag capture_log: true
+    test "should handle timeout error" do
+      %{uuid: uuid} =
+        listing =
+        insert(
+          :listing,
+          rooms: 2,
+          area: 80,
+          address: build(:address, state: "MS", city: "Mah City", street: "Mah Street"),
```
pmargreff: is this specific info necessary (state, city, and street)? Can we use only `build(:address)` here?

apps/re_integrations/lib/priceteller/price_teller.ex
rhnonose: It is in `apps/re_integration/mix.exs`

apps/re/test/price_suggestions/price_suggestions_test.exs
rhnonose: No, probably some copy&pasta duplication 😅

apps/re_integrations/lib/priceteller/price_teller.ex
pmargreff: Sorry, missed this.

PR #612, opened 2019-06-17
Expose listing uuid for admins

rhnonose:


rhnonose:
@pmargreff for now just to be used in garagem

PR #614, opened 2019-06-19
Add is_release param to dashboard active_listings_count

gabiseabra:
This allows for filtering `active_listing_count` by market type.


apps/re_web/lib/graphql/resolvers/dashboard.ex
```diff
@@ -20,6 +20,17 @@ defmodule ReWeb.Resolvers.Dashboard do
...
+     Repo.one(
+       from(
+         l in Re.Listing,
+         where: l.status == "active" and l.is_release == ^is_release,
+         select: count(l.id)
+       )
+     )}
+  end
+
   def active_listing_count(_params, _res) do
```
pmargreff: If it's only used on garagem's dashboard, this call probably can be removed.

PR #615, opened 2019-06-21
Add Salesforce+Zapier integration client

rhnonose:
To be able to replace skyvia's integration (which has some limitations), we'll be adding a zap that easily integrates with salesforce.

This code is a client for zapier's `create_salesforce_lead` webhook.

### DIFF
--- a/apps/re/test/buyer_leads/salesforce_test.exs
+++ b/apps/re/test/buyer_leads/salesforce_test.exs
@@ -1,0 +1,46 @@
+defmodule Re.BuyerLeads.SalesforceTest do
+  use Re.ModelCase
+  use Mockery
+
+  import Re.Factory
+
+  alias Re.BuyerLeads.Salesforce.Client
+
+  describe "create_lead/1" do
+    test "should send a request to create a lead" do
+      lead = insert(:buyer_lead)
+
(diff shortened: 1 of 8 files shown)


apps/re/test/buyer_leads/salesforce_test.exs
```diff
@@ -0,0 +1,46 @@
...
+      lead = insert(:buyer_lead)
+
+      mock(HTTPoison, :post, {:ok, %{status_code: 200, body: ~s({"status":"success"})}})
+
+      assert {:ok, %{"status" => "success"}} = Client.create_lead(lead)
+
+      {:ok, encoded_lead} = Jason.encode(lead)
+      uri = URI.parse("http://www.emcasa.com/salesforce_zapier")
+
+      assert_called(HTTPoison, :post, [^uri, ^encoded_lead])
```
pmargreff: Don't know if these assertions are effective, once tests will only test what occurs and not how it was done...

rhnonose: I don't understand what you mean.
Should I hard-code the encoded-lead as a strinrg and URI structure or not test at all?

rhnonose: If you want soem changes on these assertions, I can improve in the next PR as I'm doing similar tests.

PR #619, opened 2019-06-25
feature(listings): allow exclude listings from primary market who are…

pmargreff:
Allow removing listings from search when has multiple similar results. 

Now listings with the different area and features, are marked manually to appear on integrations. Someday we can change this heuristic, but the principle will be the same, don't show similar links and make search bad/ugly with multiple results.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ erl_crash.dump
 /cover
 /test/temp
 .tool-versions-e
+/.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -25,4 +25,3 @@ erl_crash.dump
 /cover
 /test/temp
 .tool-versions-e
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -25,3 +25,4 @@ erl_crash.dump
 /cover
 /test/temp
 .tool-versions-e
+/.vscode
```
rhnonose: Maybe this shouldn't be here, you should ignora in your global `.gitignore`

pmargreff:
@gabiseabra you can use this filter on queries instead filter `is_release` attribute, so we don't make primary market listings noisy.

.gitignore
pmargreff: **TIL** `git config --global core.excludesfile`

PR #622, opened 2019-06-26
Add new district status 

beligante:
I'm including a third status for districts, `partially_covered`, and renamed the `active` and `inactive` statuses to `covered` and `uncovered` to match the new business rules where a neighborhood that is partially covered can be queryable but does not generate a lead


apps/re/lib/addresses/neighborhoods.ex
```diff
@@ -34,7 +34,8 @@ defmodule Re.Addresses.Neighborhoods do
     end
   end
 
-  def districts, do: Repo.all(from(d in District, where: d.status == "active"))
+  def districts,
```
pmargreff: you can define `["partially_covered", "covered"]` to a const @whatever and give some meaningful name to it, as `public_covered` or `covered_for_show` or something that describe it.

PR #623, opened 2019-06-26
Use default values for price suggestion

rhnonose:
As a workaround for listings that are being inserted empty, I'm putting some default values to be asked to unclutter the jobs until we can figure out a better solution.


rhnonose:
Closed in favor of ignoring the job to update suggested price rather than requesting with a default value.

PR #624, opened 2019-06-27
Add moema district as a partially_covered district

beligante:


apps/re/lib/mix/tasks/addresses/add_district.ex
```diff
@@ -25,5 +25,17 @@ defmodule Mix.Tasks.Re.Addresses.AddDistrict do
       |> Repo.insert()
```
rhnonose: You should remove the district that was being added previously as it'll give some conflict errors.

pmargreff:
Added a new commit to allow to run tasks in the right way. 

@rhnonose the problem was the path, [looking to phoenix](https://github.com/phoenixframework/phoenix/tree/master/lib/mix/tasks) I found tasks should be directly under `tasks` tree. Also fixed an [initialization problem to use Ecto in tasks](https://stackoverflow.com/a/56134650). 

---

@beligante You should be able to import the new value running the task you've just changed. 

`mix compile`
`mix help` (tasks should appear under `re.addresses.add_district` name)
`mix re.addresses.add_district`

PR #635, opened 2019-07-03
Exclude duplicate developments from results

gabiseabra:
This excludes listings from developments that have already been shown on search results when `exclude_similar_for_primary_market` is true.

TODO:
- [x] Write tests
- [x] Fix remaining count query


apps/re/test/listings/listings_test.exs
```diff
@@ -282,6 +282,60 @@ defmodule Re.ListingsTest do
...
+
+      assert %{remaining_count: 1, listings: listings1} = Listings.paginated(%{
+        page_size: 2,
+        exclude_similar_for_primary_market: true
+      })
+
+      assert 2 == length(listings1)
+      listings1_ids = listings1 |> Enum.map(&Map.get(&1, :id))
+
+      assert %{remaining_count: 0, listings: listings2} = Listings.paginated(%{
```
pmargreff: this test should be split into two different cases

PR #636, opened 2019-07-03
Add typologies to development type

gabiseabra:
Add `typologies` field to graphql's `Development` type, which returns a data on listings grouped by area and rooms count.

TODO:
- [x] Write tests
- [x] Use dataloader


apps/re_web/lib/graphql/resolvers/developments.ex
```diff
@@ -38,6 +38,14 @@ defmodule ReWeb.Resolvers.Developments do
     end)
   end
 
+  def typologies(development, _params, _context) do
+    Developments.get_typologies(development.uuid)
+  rescue
```
pmargreff: I guess we could avoid the rescue, what's the situation where it can raise some problem?

PR #639, opened 2019-07-05
Add inactivation reason on listings

pmargreff:
Allow user to set a reason when deactivating listings. Also, add sold_price to keep price when a listing is deactivated cause was sold (both for us or another real estate companies). 

A workaround made to keep the history complete when it's inactivated is to keep the reason as the status (only on status story), so if a listing is deactivated but changes the reason from  `temporarely_suspended` to `sold_by_emcasa` we keep recording these changes (if the inactivation has no reason I keep `inactive` as status).  

Last but not least, by now both price and reason are not required, and validations are really soft (like allow a `sold_price` for other reasons). Things tend to be more when the demand grows up.

### DIFF
--- a/apps/re/lib/listings/schemas/listing.ex
+++ b/apps/re/lib/listings/schemas/listing.ex
@@ -45,7 +45,7 @@ defmodule Re.Listing do
     field :in_person_visit_count, :integer, virtual: true
     field :suggested_price, :float
 
-    field :inactivation_reason, :string
+    field :deactivation_reason, :string
     field :sold_price, :integer
 
     belongs_to :address, Re.Address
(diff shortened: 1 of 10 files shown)


apps/re/lib/listings/schemas/listing.ex
```diff
@@ -88,14 +91,20 @@ defmodule Re.Listing do
 
   @sun_period_types ~w(morning evening)
 
+  @inactivation_reasons ~w(duplicity exclusivity give_up left_emcasa
+                          publication_phase publication_error qualification_phase
+                          rented sold sold_by_emcasa temporarely_suspended)
```
rhnonose: There's some mixed formats here.
If it's in past tense (like in `rented`), `give_up` should be `gave_up`, `qualitifcation_phase` should be `qualified`, `publication_phase` should be `published` etc
Also, `publication_mistake` is more clear than `publication_error`.

rhnonose: Also, there's a typo, should be `temporarily_suspended`

PR #640, opened 2019-07-05
Walk-in offline buyer lead

rhnonose:
This is a new lead type for an offline walk-in marketing campaign, coming from a typeform zap.

- [x] add new buyer lead
- [x] process buyer lead
- [x] modify `buyer_lead` to contain new fields
- [x] send them to salesforce


apps/re/lib/buyer_leads/walk_in_offline.ex
```diff
@@ -0,0 +1,93 @@
...
+               where_did_you_find_about)a
+  @params @required ++ @optional
+  @locations ~w(RJ SP)
+
+  def changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, @params)
+    |> validate_required(@required)
+    |> validate_inclusion(:location, @locations,
+      message: "should be one of: [#{Enum.join(@locations, " ")}]"
```
pmargreff: We removed this messaging pattern in favor of the default from validate_required

rhnonose: I'll open a PR with the change.

PR #642, opened 2019-07-08
Remove aliases for neighborhoods

beligante:
Thir PR remove the old aliases for Vila Mariana and Pompeia, from now on, the aliases will be considered individual neighborhoods.

Depends on: https://github.com/emcasa/backend/pull/644


apps/re/lib/mix/tasks/addresses.add_district.ex
```diff
@@ -13,13 +13,16 @@ defmodule Mix.Tasks.Re.Addresses.AddDistrict do
...
   @shortdoc "Create new district"
   @partially_covered_districts [
-    "Aclimação",
-    "Bosque da Saúde",
-    "Brooklin",
-    "Cerqueira César",
-    "Chácara Inglesa",
-    "Vila Madalena",
-    "Vila Olímpia"
+    "Chácara Klabin"
```
pmargreff: Is that right? we should accept listings from this neighborhood, if we consider it a partial district, we will allow only to add new listings with garagem.

apps/re/lib/mix/tasks/addresses.add_district.ex
```diff
@@ -13,13 +13,16 @@ defmodule Mix.Tasks.Re.Addresses.AddDistrict do
...
+  ]
+
+  @covered_districts [
+    "Paraíso",
```
pmargreff: I guess you've missed Vila Pompeia here, so if we keep receiving addresses as vila pompeia we'll reject them.

PR #647, opened 2019-07-10
fix(filters): allow only one distinct on filter query

pmargreff:
Remove distinct from the individual filter and accumulate them only after creating the query.

### DIFF
--- a/apps/re/lib/listings/filters/filters.ex
+++ b/apps/re/lib/listings/filters/filters.ex
@@ -84,30 +84,17 @@ defmodule Re.Listings.Filters do
   end
 
   defp build_query(params, query) do
-    Enum.reduce(params, query, &attr_filter/2)
+    params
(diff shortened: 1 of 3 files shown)


apps/re/lib/listings/filters/filters.ex
```diff
@@ -83,7 +83,33 @@ defmodule Re.Listings.Filters do
...
+    Enum.reduce(params, query, &attr_filter/2)
+    |> apply_distinct(params)
+  end
+
+  defp apply_distinct(query, %{exclude_similar_for_primary_market: true}) do
+    from(
+      l in query,
+      distinct: coalesce(l.development_uuid, l.uuid)
+    )
+  end
```
rhnonose: You can put this all in one line using expressions: `defp apply_distinct(query, %{tags_slug: _}), do: distinct(query, [l], l.uuid)`.
Same thing for the next pattern.

PR #648, opened 2019-07-10
Update elixir and erlang verisons

rhnonose:


rhnonose:
Heroku doesn't have OTP 22.0 yet

PR #650, opened 2019-07-10
Priceteller/create liquidity indicator

pmargreff:
Allow rank listings using the difference between price and price_suggestion.


pmargreff:
@rhnonose I've added the task to populate the liquidity_ratio. 

Also probably will start a new discussion for a better structure then use all things embedded on listing schema, at the end I've found it's not the only indicator for liquidity, but only a factor ([you can get a better context here](https://www.notion.so/emcasa/48f140c47d584cf98d81528481fe5898?v=79ec81b3f13f4bdbb5e6b56876b4c9b7&p=ccea4171ee7c4eca9c5f5353a0b3521b)). But in the future I'll move this for a ListingRatio table or something like this ...

PR #652, opened 2019-07-12
feature(listings): sort listings by liquidity

pmargreff:
I'll open a new PR removing score (farol) and all features related to it.


pmargreff:
Some tests will randomly break once they're checking order against, I'm also replacing these tests in another branch, I'll only merge this PR when open the one removing score, so we don't keep tests breaking for a long time ...

PR #658, opened 2019-07-15
feature(score): replace score references with liquidity_ratio

pmargreff:
Remove score field and replace all references to score with liquidity_ratio.


pmargreff:
> Did you check if there's any client using score? Might keep it returning nil until and mark as deprecated just in case. 

Good point, I'll change it.

> Also, the 3.0 and 4.0 ratios don't make sense since ratio is around 0~1, am I correct?

Yeah, right now it can go until 1.0, I didn't find any good strategy to replace plain insertions, maybe an `insert_as_last()` and `insert_as_first()` on factory and always insert +/- 0.1 on value.

rhnonose:
SGTM.

PR #659, opened 2019-07-15
chore(tags): add new tag

pmargreff:
This tag is useful for show listings on the home page, and in the feature even in some integrations. Adding as `infrastructure` cause I don't know if we're able to add new categories and use it on garagem.

### DIFF
--- a/apps/re/lib/mix/tasks/tags.add_new.ex
+++ b/apps/re/lib/mix/tasks/tags.add_new.ex
@@ -5,7 +5,7 @@ defmodule Mix.Tasks.Re.Tags.AddNew do
   use Mix.Task
 
   @tags [
-    %{category: "infrastructure", name: "Destaque", visibility: "private"}
+    %{category: "realty", name: "Destaque", visibility: "private"}
   ]
 
   alias Re.{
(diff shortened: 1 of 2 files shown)


apps/re/lib/mix/tasks/tags.add_new.ex
```diff
@@ -0,0 +1,32 @@
...
+    Mix.Task.run("app.start")
+
+    Enum.map(@tags, &insert/1)
+  end
+
+  def insert(params) do
+    {:ok, tag} =
+      %Re.Tag{}
+      |> Tag.changeset(params)
+      |> Repo.insert(on_conflict: :nothing)
```
rhnonose: I think using `Repo.insert!` here is better, since a meaningful error will be thrown instead of a pattern mismatch in `{:ok, tag}`.

pmargreff: logging error instead, so we don't interupt the task and fail to add the tags after the error occurs...

pmargreff:
Add a soft test that ensures something was added on the database, so we don't test the logic and we need no changes every time we add a different tag...

PR #660, opened 2019-07-16
feature(filters): use distincts as subquery

pmargreff:
* Distinct on queries use to override the `order by` if we use them
as subqueries this problem is overpast.

The problem in chronological order: 
- We first [added a distinct](https://github.com/emcasa/backend/pull/635/files#diff-134765670cb29b93c4e6458e1342ab68R365) which made visible the [problem about using multiple distinct in the same query](https://sentry.io/organizations/emcasa/issues/1105898552/?project=1389074&query=is%3Aresolved). This second problem was addressed with [this PR](https://github.com/emcasa/backend/pull/647/files). 
- Now we have another problem, [a distinct will always override the leftmost order_by](https://stackoverflow.com/questions/9795660/postgresql-distinct-on-with-different-order-by) (postgresql ¯\\_(ツ)_/¯), after start to use `liquidity_ratio` to sort the problem became really visible.

For override this problem right now, I'm isolating the distinct in subqueries and getting the result with a join. I'll need to revert the new way to count listings to the old one. 

This problem should be back in the future, start to think about new approaches to search/filter is a good ideia.


apps/re/test/listings/filters/filters_test.exs
```diff
@@ -839,6 +841,33 @@ defmodule Re.Listings.FiltersTest do
...
+        |> Listings.upsert_tags([tag_1.uuid])
+
+      filters = %{
+        tags_slug: ["tag-1"]
+      }
+
+      result =
+        Listing
+        |> order_by([l], desc: l.id)
+        |> Filters.apply(filters)
```
rhnonose: Isn't it better to assert if the correct `order_by` was inserted in the query?
This test might be returning a false-positive.

pmargreff: Don't know if look into the query is a good thing, will check how we do something and not the behavior. [Postgres append an 'asc'](https://www.postgresql.org/docs/9.1/queries-order.html) by default on order by, so it will only be falsely positive if they change in.

I'll skip it right now, but if I came back to this problem again will change the test approach. Looks reasonable to you?

apps/re/lib/listings/filters/filters.ex
```diff
@@ -249,8 +249,8 @@ defmodule Re.Listings.Filters do
 
   defp attr_filter({:tags_slug, slugs}, query) do
     listings_tags =
-      from(
-        l in query,
```
rhnonose: I prefer the expression form when using with pipes:
```
query
|> join([l], t in assoc(l, :tags)
|> where([l, t], t.name_slug in ^slugs)
|> distinct([l, _], l.uuid)
```
(or at least I think it's like this, not sure)

apps/re/test/listings/filters/filters_test.exs
rhnonose: Yep.

apps/re/lib/listings/filters/filters.ex
pmargreff: good catch!

pmargreff:
We've made some changes since the first approach, basically changed the complex subqueries to use a simple `group_by id`, instead of a `distinct`, so all query with order by will work, at the same time it doesn't return duplicated results.

PR #662, opened 2019-07-16
Price suggestion request lead

rhnonose:
- [x] Create a `seller_lead` entity
- [x] Add `uuid` to `price_suggestion_requests` and `addresses`
- [ ] Create processing job upon inserting `price_suggestion_request`
- [ ] Process request into `seller_lead`
- [ ] Send to salesforce through zapier client

This PR should not change any behavior.


apps/re/lib/seller_leads/seller_lead.ex
```diff
@@ -0,0 +1,51 @@
...
+    field :source, :string
+    field :complement, :string
+    field :type, :string
+    field :area, :string
+    field :maintenance_fee, :float
+    field :rooms, :integer
+    field :bathrooms, :integer
+    field :suites, :integer
+    field :garage_spots, :integer
+    field :value, :string
```
pmargreff: shouldn't we use price here? so we keep the same vocabulary for both lead and listings...

rhnonose: I mapped the named used on the lead itself, but you might be right.

rhnonose: Aaand it's also a string, will change to `float` since it's the type set in `price_suggestion_request`.
We need some consistency on prices 😅

PR #663, opened 2019-07-17
Routific integration

gabiseabra:
Integration with routific api and google calendars to generate photographers' schedules. This PR covers calls to routific api.


apps/re_integrations/test/routific/routific_test.exs
```diff
@@ -0,0 +1,37 @@
...
+
+  describe "start_job/1" do
+    test "create a new job to monitor routific request" do
+      assert {:ok, _} = Routific.start_job(@visits)
+      assert_enqueued_job(Repo.all(JobQueue), "monitor_routific_job")
+    end
+  end
+
+  describe "get_job_status/1" do
+    test "fetch routific job status" do
```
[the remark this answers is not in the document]
gabiseabra: This PR only covers routific api calls. The actual job queue flow, which interfaces with google calendars, will be implemented in another PR.

apps/re_integrations/lib/routific/job_queue.ex
```diff
@@ -0,0 +1,24 @@
+defmodule ReIntegrations.Routific.JobQueue do
+  @moduledoc """
+  Module for processing jobs related with routific payloads.
+  """
+
+  use EctoJob.JobQueue, table_name: "routific_jobs", schema_prefix: "re_integrations"
```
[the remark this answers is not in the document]
rhnonose: If this PR is merged as I suggested, it's better to keep this deactivated.

apps/re_integrations/lib/routific/payload/inbound.ex
```diff
@@ -0,0 +1,38 @@
...
+      address: visit["location_name"],
+      start: Map.get(visit, "arrival_time"),
+      end: Map.get(visit, "finish_time")
+    }
+  end
+
+  defp build_unserved(nil), do: []
+
+  defp build_unserved(unserved) do
+    Enum.map(unserved, fn {lead_id, reason} -> {lead_id, reason} end)
```
rhnonose: This line seems to be doing nothing, what's the purpose?

apps/re_integrations/lib/routific/routific.ex
```diff
@@ -0,0 +1,33 @@
...
+      |> Repo.insert()
+    end
+  end
+
+  def get_job_status(job_id) do
+    with {:ok, payload = %{status: "finished"}} <- Client.fetch_job(job_id) do
+      {:ok, payload}
+    else
+      {_, %{status_code: status_code}} -> {:error, %{status_code: status_code}}
+      {_, %{status: status}} -> {:error, %{status: status}}
```
rhnonose: I'm not sure here if you should handle expected status with the `{:error, _}` tuple.
The API call succeeds and it returns the job status, so returning as an error gives the wrong signal.

apps/re_integrations/lib/routific/job_queue.ex
```diff
@@ -0,0 +1,22 @@
...
+
+  alias ReIntegrations.{
+    Repo,
+    Routific
+  }
+
+  alias Ecto.Multi
+
+  def perform(%Multi{} = multi, %{"type" => "monitor_routific_job", "job_id" => id}) do
```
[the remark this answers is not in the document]
gabiseabra: I found ecto_job to be fitting, although not ideal, since it limits max retries per job, unlike most cron libs which run tasks periodically, and the backoff is actually linear (should be `reservation_timeout` + `execution_timeout`). I agree there's a batter solution for this, though. Should I rewrite this step using a `Task.Supervisor`?

apps/re_integrations/lib/routific/payload/inbound.ex
gabiseabra: This is supposed to turn a map into a list of tuples.

apps/re_integrations/lib/routific/job_queue.ex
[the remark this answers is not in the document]
rhnonose: A `Task` won't survive a restart, so we can have a pending restart.
If we persist a scheduling job with a state, a cron-job can check for jobs that are waiting for a response from `routific` and will survive restarts.
But that "job" shouldn't be an `ecto_job` since we can't control job states.
Let's talk about it later today.

apps/re_integrations/lib/routific/payload/inbound.ex
rhnonose: But why do you need a list of tuples instead of a map?
If you use `Enum`, you won't need to convert like that.

apps/re_integrations/lib/routific/routific.ex
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  @shift Application.get_env(:re_integrations, :routific_shift, {~T[08:00:00Z], ~T[18:00:00Z]})
+  @max_attempts Application.get_env(:re_integrations, :routific_max_attempts, 6)
+
+  def shift_start, do: elem(@shift, 0)
+  def shift_end, do: elem(@shift, 1)
+
+  def start_job(visits) do
+    with {:ok, %{body: body}} <- visits |> Payload.Outbound.build() |> Client.start_job(),
```
rhnonose: `Payload.Outbound.build()` can be it's own `with` clause, returning `{:ok, payload}`.
Otherwise, if I pass an invalid `visits`, it'll break on the client.

apps/re_integrations/lib/routific/routific.ex
```diff
@@ -0,0 +1,43 @@
...
+         {:ok, payload = %{"status" => "finished"}} <- Jason.decode(body) do
+      {:ok, Payload.Inbound.build(payload)}
+    else
+      {:ok, data = %{status_code: status_code}} when status_code != 200 ->
+        {:error, data}
+
+      {:ok, payload = %{"status" => status}} ->
+        {String.to_atom(status), Payload.Inbound.build(payload)}
+
+      {:error, data} ->
```
rhnonose: This could be a default clause `error -> error`.

apps/re_integrations/lib/routific/routific.ex
```diff
@@ -0,0 +1,43 @@
...
+  def get_job_status(job_id) do
+    with {:ok, %{body: body}} <- Client.fetch_job(job_id),
+         {:ok, payload = %{"status" => "finished"}} <- Jason.decode(body) do
+      {:ok, Payload.Inbound.build(payload)}
+    else
+      {:ok, data = %{status_code: status_code}} when status_code != 200 ->
+        {:error, data}
+
+      {:ok, payload = %{"status" => status}} ->
+        {String.to_atom(status), Payload.Inbound.build(payload)}
```
rhnonose: Converting responses from an external API to atom is a no-no.
It's better if handle them as a string.

gabiseabra:
I used errors as an alternative to building all visits and then looping through the resulting list to find invalid items, but this is a valid point. I agree that using exceptions for control flow isn't great since they're not meant for that.
I didn't realize ecto's chagesets could be used to validate arbitrary structs, I'll consider it next time. Thanks for the tips and criticism.

rhnonose:
Thanks for sticking through a lot of changes.
There's still a few things to improve, but I'll invite you to a refactoring instead of going through too much on this PR.

PR #664, opened 2019-07-17
Google calendar integration

gabiseabra:
This PR introduces the `calendar` entity to store photographers' shifts and calendar ids and handling events on those calendars with google's api.
TODO:
- [x] Test google calendars integration


apps/re/lib/google_calendars/calendar/queries.ex
```diff
@@ -0,0 +1,29 @@
...
+  import Ecto.Query
+
+  @relations [:districts]
+
+  def preload_relations(query \\ Calendar, relations \\ @relations)
+
+  def preload_relations(query, relations), do: preload(query, ^relations)
+
+  def by_districts(query \\ Calendar, district_names_or_slugs) do
+    district_slugs = district_names_or_slugs |> Enum.map(&Slugs.sluggify/1)
```
rhnonose: The way we implemented the `sluggify` function, sluggifying again will create an invalid slug, so it's better if you always pass a slug and then remove this sluggify transformation.

apps/re/lib/google_calendars/calendar_district.ex
```diff
@@ -0,0 +1,32 @@
...
+  schema "calendar_districts" do
+    belongs_to :listing, Re.GoogleCalendars.Calendar,
+      type: :binary_id,
+      foreign_key: :calendar_uuid,
+      references: :uuid,
+      primary_key: true
+
+    belongs_to :tag, Re.Addresses.District,
+      type: :id,
+      foreign_key: :district_id,
```
rhnonose: If there's no `uuid` for district, we should generate it.
See [this migration](apps/re/priv/repo/migrations/20190716185939_generate_address_uuid.exs).

apps/re/test/google_calendars/calendar/queries_test.exs
```diff
@@ -0,0 +1,31 @@
...
+  import Re.Factory
+
+  describe "by_districts/1" do
+    test "should get all calendars by districts name" do
+      district1 = insert(:district)
+      district2 = insert(:district)
+      insert(:calendar, districts: [district1])
+
+      calendars =
+        Enum.sort([
```
rhnonose: If order is not important for this asseriton, we have a custom one: `assert_mapper_match` in `Re.CustomAssertion`

apps/re_integrations/mix.exs
```diff
@@ -42,15 +42,17 @@ defmodule ReIntegrations.Mixfile do
...
       {:kadabra, "~> 0.4.2"},
       {:retry, "~> 0.10"},
       {:httpoison, "~> 1.3", override: true},
       {:jason, "~> 1.0"},
       {:sentry, "~> 7.0"},
       {:ex_machina, "~> 2.2", only: :test},
       {:ecto_job, "~> 2.0"},
-      {:cloudex, git: "https://github.com/emcasa/cloudex.git"}
+      {:cloudex, git: "https://github.com/emcasa/cloudex.git"},
+      {:google_api_calendar, "~> 0.0.1"},
```
rhnonose: This API is on verstion `0.6`, we should get the latest when possible.

apps/re_integrations/lib/google/calendars/calendars.ex
```diff
@@ -0,0 +1,76 @@
...
+    GoogleCalendars,
+    GoogleCalendars.Calendar
+  }
+
+  @token Application.get_env(:re_integrations, :goth_token, Goth.Token)
+  @timezone Application.get_env(:re_integrations, :google_api_timezone, "America/Sao_Paulo")
+  @acl_owner Application.get_env(:re_integrations, :google_calendar_acl_owner, nil)
+
+  @default_event_opts [timeZone: @timezone]
+  @default_acl if @acl_owner,
```
rhnonose: I don't feel comfortable having conditionals in module attributes. This seems hard to debug if there's a problem.
You can build the struct in the config file and return here, removing this conditional (and the other one in `insert_acl/1`.
This makes for a more explicit configuration when reading the config file itself.

config/prod.exs
```diff
@@ -85,7 +85,13 @@ config :re_integrations,
...
   orulo_url: System.get_env("ORULO_URL"),
   orulo_api_token: System.get_env("ORULO_API_TOKEN"),
-  orulo_client_token: System.get_env("ORULO_CLIENT_TOKEN")
+  orulo_client_token: System.get_env("ORULO_CLIENT_TOKEN"),
+  google_calendar_acl_owner:
+    with(
+      acl_type when not is_nil(acl_type) <- System.get_env("GOOGLE_CALENDAR_ACL_OWNER_TYPE"),
+      acl_owner when not is_nil(acl_owner) <- System.get_env("GOOGLE_CALENDAR_ACL_OWNER"),
+      do: [type: acl_type, value: acl_owner]
+    )
```
rhnonose: As I referred in another comment, these conditionals should not be here.
We can assume those env_vars will be filled and calling `Application.get_env/3` should handle `nil` values.

config/prod.exs
```diff
@@ -112,3 +118,7 @@ config :cloudex,
   api_key: System.get_env("CLOUDINARY_API_KEY"),
   secret: System.get_env("CLOUDINARY_SECRET"),
   cloud_name: System.get_env("CLOUDINARY_CLOUD_NAME")
+
+config :goth,
+  json: {:system, "GCP_CREDENTIALS"},
+  disabled: is_nil(System.get_env("GCP_CREDENTIALS"))
```
rhnonose: Same thing for nil env var comment before.

apps/re/lib/google_calendars/calendar/calendar.ex
```diff
@@ -0,0 +1,42 @@
...
+    many_to_many :districts, Re.Addresses.District,
+      join_through: Re.GoogleCalendars.CalendarDistrict,
+      join_keys: [calendar_uuid: :uuid, district_uuid: :uuid],
+      on_replace: :delete
+
+    timestamps()
+  end
+
+  @required ~w(external_id shift_start shift_end)a
+  @optional ~w()a
```
pmargreff: no need an optional here once they do not exists ...

apps/re_integrations/lib/google/calendars/calendars.ex
```diff
@@ -0,0 +1,67 @@
...
+  @default_acl Application.get_env(:re_integrations, :google_calendar_acl)
+
+  @default_event_opts [timeZone: @timezone]
+
+  defp conn do
+    {:ok, token} = @token.for_scope("https://www.googleapis.com/auth/calendar")
+    Connection.new(token.token)
+  end
+
+  def get_events(%Calendar{} = calendar, opts \\ []) do
```
pmargreff: Will we need this? I thought at the first version we will ignore the current activities from the schedule...

PR #665, opened 2019-07-17
Zapier salesforce seller client

rhnonose:
Following up on #662 to add a client to send seller leads to salesforce via zapier
The client configuration is pretty similar to buyer leads, but for now for simplicity we'll keep it separated.
When there's a need to extract some stuff, we'll do.

### DIFF
--- a/apps/re/priv/repo/migrations/20190717190328_fix_area_type.exs
+++ b/apps/re/priv/repo/migrations/20190717190328_fix_area_type.exs
@@ -1,0 +1,25 @@
+defmodule Re.Repo.Migrations.FixAreaType do
+  use Ecto.Migration
+
+  def up do
+    alter table(:seller_leads) do
+      remove :area
+    end
+
+    flush()
+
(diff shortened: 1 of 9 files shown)


apps/re/priv/repo/migrations/20190717190328_fix_area_type.exs
```diff
@@ -0,0 +1,25 @@
+defmodule Re.Repo.Migrations.FixAreaType do
+  use Ecto.Migration
+
+  def up do
+    alter table(:seller_leads) do
```
pmargreff: no need to changes, but you could use [modify](https://hexdocs.pm/ecto_sql/Ecto.Migration.html?#modify/3) here.

rhnonose: I didn't want to have to cast the types.
Since we're not using it in production yet, it doesn't matter.

PR #669, opened 2019-07-19
add new column for user type and salesforce id

beligante:
… integrations


apps/re/priv/repo/migrations/20190719123600_create_user_type_concept.exs
```diff
@@ -0,0 +1,19 @@
+defmodule Re.Repo.Migrations.CreateUserTypeConcept do
+  use Ecto.Migration
+
+  def up do
+    alter table(:users) do
+      add :type, :string, default: "common"
```
pmargreff: We don't need to set a default on both database and application, do it only in the application is good enough to me.

apps/re/lib/accounts/user.ex
```diff
@@ -16,6 +16,9 @@ defmodule Re.User do
     field :device_token, :string
     field :account_kit_id, :string
 
+    field :type, :string, default: "common"
+    field :sales_force_id, :string
```
rhnonose: We're keeping `salesforce` together in most places.

apps/re/priv/repo/migrations/20190719123600_create_user_type_concept.exs
```diff
@@ -0,0 +1,19 @@
...
+      add :sales_force_id, :string
```
rhnonose: Please add a unique constraint for this.

apps/re/priv/repo/migrations/20190719123600_create_user_type_concept.exs
```diff
@@ -0,0 +1,28 @@
...
+    create unique_index(:users, [:salesforce_id])
+
+    flush()
+
+    q = from(u in "users")
+    Re.Repo.update_all(q, set: [type: "common"])
+  end
+
+  def down do
+    drop_if_exists unique_index(:users, [:salesforce_id])
```
rhnonose: This is not required as it'll be automatically dropped when you remove the column.

apps/re/lib/accounts/user.ex
```diff
@@ -31,9 +34,10 @@ defmodule Re.User do
   end
 
   @roles ~w(admin user)
+  @types ~w(common partner_broker)
```
rhnonose: I don't like `common`.
Should reflect our expected user type which is `property_owner` or `owner`.

beligante: I thought about that, but what I'm calling `common` user may not be a `property owner`

beligante: but the average users will be property owners, thinking better, this change make sense

apps/re/priv/repo/migrations/20190719123600_create_user_type_concept.exs
```diff
@@ -0,0 +1,26 @@
...
+    flush()
+
+    q = from(u in "users")
+    Re.Repo.update_all(q, set: [type: "property_owner"])
+  end
+
+  def down do
+    alter table(:users) do
+      remove :type
+      remove :sales_force_id
```
pmargreff:
```suggestion
      remove :salesforce_id
```

apps/re/lib/accounts/user.ex
rhnonose: It's because of the workflow we're implementing.
The interface will assume between owner and partner broker and will be treated as such.
If they do not belong to either type of user, we don't have anything for them in our product anyway, so there's no reason to "map" them for now.

PR #671, opened 2019-07-22
Send price suggestion salesforce

rhnonose:
Follow-up modification to complement #662 and #665
- Job to process `price_suggestion_request` to `seller_lead`
- Job to send `seller_lead` to salesforce through zapier

~Pending manual test to check zapier is fine~

### DIFF
--- a/apps/re/lib/price_suggestions/price_suggestions.ex
+++ b/apps/re/lib/price_suggestions/price_suggestions.ex
@@ -77,22 +77,10 @@ defmodule Re.PriceSuggestions do
   defp preload_address(params), do: params
 
   def create_request(params) do
-    with %{valid?: true} = changeset <- Request.changeset(%Request{}, params),
+    with changeset <- Request.changeset(%Request{}, params),
          suggested_price <- do_suggest_price(params),
-         %{valid?: true} = changeset <- set_suggested_price(changeset, suggested_price) do
-      uuid = Changeset.get_field(changeset, :uuid)
-
-      Ecto.Multi.new()
-      |> Multi.insert(:insert_price_suggestion_request, changeset)
-      |> JobQueue.enqueue(:seller_lead_job, %{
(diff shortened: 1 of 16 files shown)


apps/re/lib/price_suggestions/price_suggestions.ex
```diff
@@ -82,30 +72,39 @@ defmodule Re.PriceSuggestions do
...
 
   defp preload_address(params), do: params
 
-  def create_request(params, %{id: address_id}, user) do
-    %Request{}
-    |> Changeset.change(address_id: address_id)
-    |> attach_user(user)
-    |> Request.changeset(params)
-    |> Repo.insert()
+  def create_request(params) do
```
pmargreff: this function is big and doing a lot of things, extract at least the queue creation could improve readability

apps/re/test/seller_leads/job_queue_test.exs
```diff
@@ -0,0 +1,147 @@
...
+
+  alias Ecto.Multi
+
+  describe "price_suggestion_request" do
+    test "process lead" do
+      %{uuid: address_uuid} = address = insert(:address)
+      %{uuid: user_uuid} = user = insert(:user)
+
+      %{uuid: uuid} =
+        lead =
```
pmargreff: maybe `unprocessed_leads` or `price_suggestion` as name here? The assertions `assert seller... == lead...` are confusing while reading.

apps/re/test/interests/interests_test.exs
```diff
@@ -50,13 +56,14 @@ defmodule Re.InterestsTest do
         is_covered: true
       }
 
-      assert {:ok, %{suggested_price: 26_279.0}} = Interests.request_price_suggestion(params, nil)
```
pmargreff: a `requist_price_suggestion` with `nil` user isn't expected anymore? If it's duplicate the test is a better idea.

PR #673, opened 2019-07-22
Allow listing filters with multiple values

lucas-tulio:
Added `rooms`, `bathrooms`, `suites` and `garage_spots` filters to the `listings` query. These filters allow an array of integers as parameter, so the user can select multiple values.


lucas-tulio:
Good point. We're going to keep the same range behavior and make changes in the UI only, to make things simpler 👍

PR #675, opened 2019-07-23
fix(leads): allow leads with more than 255 chars on url

pmargreff:
Fixes [BACKEND-1P](https://sentry.io/organizations/emcasa/issues/1074544642/?project=1389074)

### DIFF
--- a/apps/re/priv/repo/migrations/20190723134327_modify_buyer_lead_url_size.exs
+++ b/apps/re/priv/repo/migrations/20190723134327_modify_buyer_lead_url_size.exs
@@ -1,9 +1,15 @@
 defmodule Re.Repo.Migrations.ModifyBuyerLeadUrlSize do
   use Ecto.Migration
 
-  def change do
+  def up do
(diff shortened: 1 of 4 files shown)


apps/re/priv/repo/migrations/20190723134327_modify_buyer_lead_url_size.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Re.Repo.Migrations.ModifyBuyerLeadUrlSize do
+  use Ecto.Migration
+
+  def change do
```
rhnonose: Since type change doesn't save the prior type, we need a `up` and `down` to allow rollback.

pmargreff:
@lucas-tulio it's a good idea to dig into it and try to relate (solve/handle) errors on this workflow.

pmargreff:
Add requested changes, also replicate solution to `empty_search_buyer_leads` once this value is also used in this table.

PR #676, opened 2019-07-23
feat(calendars) add address to calendar table

gabiseabra:
Add address id to calendars to act as a starting point for vehicle routing.


apps/re/lib/google_calendars/calendar/calendar.ex
```diff
@@ -18,6 +18,8 @@ defmodule Re.GoogleCalendars.Calendar do
       join_keys: [calendar_uuid: :uuid, district_uuid: :uuid],
       on_replace: :delete
 
+    belongs_to :address, Re.Address
```
rhnonose: Please make this reference through `uuid`.

apps/re/priv/repo/migrations/20190719214009_add_calendars_depot.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Re.Repo.Migrations.AddCalendarsDepot do
+  use Ecto.Migration
+
+  def change do
+    alter table(:calendars) do
+      add :address_id, references(:addresses)
```
rhnonose: Same as in schema, make this reference through `uuid`.

PR #677, opened 2019-07-23
Partner broker

beligante:


apps/re/priv/repo/migrations/20190723164744_create_partner_broker_relation.exs
```diff
@@ -2,7 +2,7 @@ defmodule Re.Repo.Migrations.CreatePartnerBrokerRelation do
   use Ecto.Migration
 
   def change do
-    create table(:broker_district, primary_key: false) do
+    create table(:broker_districts, primary_key: false) do
```
rhnonose: should be `brokers_districts` since it's a `many_to_many` relationship.

apps/re_web/lib/graphql/types/user.ex
```diff
@@ -127,5 +127,17 @@ defmodule ReWeb.Types.User do
 
       resolve &AccountsResolver.change_role/2
     end
+
+    @desc "Edits and user so it became a partner broker"
+    field :edit_partner_broker, type: :user do
```
rhnonose: Not sure if we need a new mutation for this, we already have `edit_user_profile`.

apps/re_web/lib/graphql/types/user.ex
```diff
@@ -127,5 +127,17 @@ defmodule ReWeb.Types.User do
...
+      arg :id, non_null(:id)
+      arg :name, :string
+      arg :phone, :string
+      arg :device_token, :string
+      arg :notification_preferences, :notification_preferences_input
+      arg :districts, non_null(list_of(:string))
```
rhnonose: This shouldn't be a list of strings.
Now that we're going to use districts back, we should return an ID through the district query and then use it here.

apps/re/lib/accounts/users.ex
```diff
@@ -41,20 +41,19 @@ defmodule Re.Accounts.Users do
...
-    Enum.reduce(opts, user_changeset, fn
-      {:districts, districts}, changeset ->
-        repo_districts = Neighborhoods.districts_by_name_slugs(districts)
-        Changeset.put_assoc(changeset, :districts, repo_districts)
-    end)
+  def update_districts(user, districts) do
+    repo_districts = Neighborhoods.districts_by_uuids(districts)
+    user
+    |> Repo.preload(:districts)
+    |> Changeset.change
```
rhnonose: I'm pretty sure you can just delete this line, the first argument of `Changeset` functions accept both changesets and structs AFAIK.

beligante: According with the documentation the `Changeset.put_assoc/4` only accepts changesets

PR #678, opened 2019-07-23
chore(tags): add new tags and new tag category

pmargreff:
Add new tags to give some visibility to ops about this information. I need to add the new group (`ops`) on garagem to show it nicely.


pmargreff:
Also added an entry as an example to the service account config, I could not run it locally without a valid entry.

PR #679, opened 2019-07-24
New related listings (aliketeller)

rhnonose:
This PR integrates a json dump of related listings generated by the data team.
It loads on a `GenServer` startup and daily via cron+cast into an ETS table.
If there's new listings/updated listings, it'll replace seamlessly.

Approaches considered for storing the `uuid`s:
- Agent and GenServer's state: storing on a process state would create a read bottleneck for retrieving data, not only on a synchronous call, but also loading the whole state and getting as a key
- Postgres: putting on the database would be slower and a little bit boilerplate-y. It removes the necessity to load on startup but requires migrations, schemas and queries. Since there's a lot of incoming data (a listing have 50+ related listings), it would create a ton of relations in the database which might not be good if we're going to switch to an API later
- `ets` seems like a better alternative since it eases that bottleneck and the upsert logic used is out-of-the-box.

- [x] Load related listings on startup
- [x] Load related listing through `GenServer.cast/2`
- [x] Interface to get related listings through `uuid`
- [x] Cronjob to load periodically
- [x] Tests
- [x] Replace old `Related` logic to use aliketeller


config/dev.secret-example.exs
```diff
@@ -50,7 +50,8 @@ config :re,
   garagem_url: "localhost",
   zapier_create_salesforce_lead_url: "SALESFORCE_ZAPIER_URL",
   priceteller_url: "priceteller_url",
-  priceteller_token: "mahtoken"
+  priceteller_token: "mahtoken",
+  aliketeller_url: "aliketeller_url"
```
pmargreff: This config is not replicated to production env, is this right?

apps/re/lib/aliketeller/server.ex
```diff
@@ -0,0 +1,59 @@
...
+    |> http_client().get()
+  end
+
+  defp save_on_ets(%{"data" => data}), do: Enum.each(data, &do_save_on_ets/1)
+
+  defp do_save_on_ets(%{"listing_uuid" => uuid, "suggested_listing_uuids" => uuids}) do
+    :ets.insert(:aliketeller, {uuid, uuids})
+  end
+
+  defp create_ets_table do
```
pmargreff: This function (maybe the related function as well) can be extracted to its own module, so we can reuse this function on tests setup.

apps/re/test/listings/related_test.exs
```diff
@@ -6,143 +6,22 @@ defmodule Re.RelatedTest do
...
-          price: 74_999
-        )
+      :ets.new(:aliketeller, [:set, :protected, :named_table, read_concurrency: true])
+      :ets.insert(:aliketeller, {listing.uuid, [related1.uuid, related2.uuid]})
 
-      assert %{listings: [%{id: ^id2}], remaining_count: 0} = Related.get(listing)
+      {:ok, listing: listing, related1: related1, related2: related2}
     end
 
-    test "should return a neighborhood and price match listing" do
```
pmargreff: Odin bless you.

apps/re_web/test/graphql/listings/query_test.exs
```diff
@@ -14,6 +14,8 @@ defmodule ReWeb.GraphQL.Listings.QueryTest do
     admin_user = insert(:user, email: "admin@email.com", role: "admin")
     user_user = insert(:user, email: "user@email.com", role: "user")
 
+    :ets.new(:aliketeller, [:set, :protected, :named_table, read_concurrency: true])
```
pmargreff: This could be extracted to a common function, or even a helper to decrease the necessity to have this logic spreadly in code.

config/dev.secret-example.exs
rhnonose: My bad, it's not.

apps/re_web/test/graphql/listings/query_test.exs
rhnonose: Dully noted.

rhnonose:
`aliketeller` is the service's name (similar to `priceteller`), I think it's ok to have it as a "provider of related listings" since there's the `related` attribute in the API and `Related` module to make the call to the API.

PR #680, opened 2019-07-25
Routific + google calendar integration

gabiseabra:
* Create google calendar events from routific job results
* Generate routific's fleet data from calendars


gabiseabra:
This PR has been closed due to a change in how we handle tour scheduling. Rather than creating google calendar events for each visit they'll be registered as a salesforce event, which in turn is added to the event owner's schedule.
This is a better approach since it obviates the need to replicate these entries to salesforce manually, and calendar events generated by salesforce are automatically synced on updates.

PR #683, opened 2019-07-29
Salesforce integration

gabiseabra:
Context for querying `emcasa/salesforce`'s api for pending tour visits.


apps/re_integrations/lib/salesforce/payload/opportunity.ex
```diff
@@ -0,0 +1,78 @@
...
+        :salesforce_morning_period,
+        "Manhã: 09h - 12h"
+      ),
+    afternoon:
+      Application.get_env(
+        :re_integrations,
+        :salesforce_afternoon_period,
+        "Tarde: 12h - 18h"
+      )
+  )
```
rhnonose: Why do those periods need to be a configuration?

apps/re_integrations/lib/salesforce/payload/opportunity.ex
```diff
@@ -0,0 +1,78 @@
...
+  def validate(params) do
+    %__MODULE__{}
+    |> changeset(params)
+    |> case do
+      %{valid?: true} = changeset -> {:ok, apply_changes(changeset)}
+      changeset -> {:error, :invalid_input, params, changeset}
+    end
+  end
+
+  defp changeset(struct, params), do: struct |> cast(params, @params)
```
rhnonose: This doesn't have to be a pipe since there's only one function call.

apps/re_integrations/lib/salesforce/salesforce.ex
```diff
@@ -0,0 +1,65 @@
...
+  Context module for salesforce
+  """
+
+  alias ReIntegrations.{
+    Routific,
+    Salesforce.Client,
+    Salesforce.Payload
+  }
+
+  @schedule_interval Application.get_env(:re_integrations, :tour_schedule_interval, days: 2)
```
rhnonose: How are we thinking about handling holidays and weekends?

apps/re_integrations/test/salesforce/payload/opportunity_test.exs
```diff
@@ -0,0 +1,48 @@
...
+      assert opportunity.address == @payload["Dados_do_Imovel_para_Venda__c"]
+      assert opportunity.neighborhood == @payload["Bairro__c"]
+      assert opportunity.tour_date == ~N[2019-07-29 20:00:00Z]
+      assert opportunity.tour_period == :morning
+    end
+  end
+
+  describe "visitation_period/1" do
+    test "returns time range from opportunity's tour_date" do
+      assert %{start: ~T[20:00:00Z], end: ~T[20:00:00Z]} =
```
rhnonose: This is a weird test, `start` and `end` are the same time.
If this is intended, the test description should be more clear about why that happens.

apps/re_integrations/lib/salesforce/salesforce.ex
gabiseabra: That's right, I should handle this in the scheduler. Thanks.

apps/re_integrations/test/salesforce/payload/opportunity_test.exs
rhnonose: Hmm, still didn't explain to me.
`start` and `end` at the same time doesn't make sense because it's an appointment that takes time, isn't it?
Shouldn't it have a minimum duration?

gabiseabra: This is a range of hours that photographers can arrive at the location to start the session. Perhaps `visit_hours` might be a better name for this function, do you have any suggestions?

rhnonose: `visit_start_window` is maybe a better name for the function and `deadline` instead of `end`.

PR #684, opened 2019-07-29
feat(routific) build fleet from calendars

gabiseabra:


apps/re_integrations/lib/routific/payload/outbound.ex
```diff
@@ -45,11 +46,39 @@ defmodule ReIntegrations.Routific.Payload.Outbound do
 
   defp build_visit(_visit), do: :error
 
-  defp build_fleet(_visits) do
-    # TODO build fleet from photographers' google calendars
-    {:ok, %{}}
+  defp build_fleet(visits) do
+    {:ok, visits
```
rhnonose: If you're expecting no errors here, the `{:ok, result}` tuple is unnecessary.
Maybe you should add a `case` statement at the end when `get_calendars/1` return an empty list.

apps/re_integrations/lib/routific/payload/outbound.ex
```diff
@@ -45,11 +46,39 @@ defmodule ReIntegrations.Routific.Payload.Outbound do
...
+  end
+
+  defp get_calendars(neighborhoods) do
+    Calendar
+    |> Calendar.Queries.by_district_names(neighborhoods)
+    |> Calendar.Queries.preload_relations([:address])
+    |> Re.Repo.all()
+  end
+
+  defp build_depot(%{address: address}), do: %{
```
rhnonose: Did you mean `depart`? I'm not sure what `depot` would mean in this context, so just guessing 😄

gabiseabra: This is routific's terminology for a vehicle's starting point.

apps/re_integrations/lib/routific/payload/outbound.ex
```diff
@@ -45,11 +46,40 @@ defmodule ReIntegrations.Routific.Payload.Outbound do
 
   defp build_visit(_visit), do: :error
 
-  defp build_fleet(_visits) do
-    # TODO build fleet from photographers' google calendars
-    {:ok, %{}}
+  defp build_fleet(visits) do
+    with {:ok, calendars} <- visits |> get_neighborhoods() |> get_calendars() do
```
pmargreff: You're not threating errors on `with`, use it without `with` as 
```
{:ok, _} = a 
|> foo()
|> bar()
|> baz()
```
could be good enough.

rhnonose: Or just flat remove `{:ok, _} = a`

PR #687, opened 2019-07-30
chore(tags): add new tags

pmargreff:


apps/re/lib/mix/tasks/tags.add_new.ex
```diff
@@ -5,10 +5,17 @@ defmodule Mix.Tasks.Re.Tags.AddNew do
...
 
   @tags [
-    %{category: "ops", name: "Temos Chave", visibility: "private"},
-    %{category: "ops", name: "Desocupado", visibility: "private"},
-    %{category: "ops", name: "Placa De Contato", visibility: "private"},
-    %{category: "ops", name: "Exclusivo", visibility: "private"}
+    %{category: "realty", name: "Armário Na Cozinha", visibility: "public"},
+    %{category: "realty", name: "Armário No Quarto", visibility: "public"},
+    %{category: "realty", name: "Área De Serviço", visibility: "public"},
+    %{category: "realty", name: "Primeira Quadra Da Praia", visibility: "public"},
```
rhnonose: This seems like a `location` category tag.

pmargreff: I don't saw this one

PR #693, opened 2019-08-01
Broker seller lead

beligante:
- [x] create mutation
- [x] insert (if doesn't exist) the property_owner from the lead (for association with seller_lead)
- [x] create tests for the mutation


apps/re_web/test/graphql/seller_leads/mutation_test.exs
```diff
@@ -139,4 +145,184 @@ defmodule ReWeb.GraphQL.SellerLeads.MutationTest do
...
+          "type" => "Apartamento"
+        }
+      }
+      conn =
+        post(conn, "/graphql_api", AbsintheHelpers.mutation_wrapper(@create_mutation, variables))
+
+      assert json_response(conn, 200)["errors"]
+    end
+
+    test "broker lead should add owner as property_owner user when it doesn't exists", %{
```
pmargreff: These test looks like a unit test to me, 'cause you're testing logic, not access neither format. You can write this test directly against `seller_leads` module. 

Also, you would write a test when the user already exists, to ensure everything works as expected.

beligante:
Sorry for the huge PR, next time I Will make sure that things stay small

apps/re/lib/accounts/user.ex
```diff
@@ -27,6 +27,8 @@ defmodule Re.User do
 
     has_many :listings, Re.Listing
 
+    has_many(:broker_leads, Re.SellerLeads.Broker, foreign_key: :broker_uuid, references: :uuid)
```
rhnonose: Please remove the enclosing `()` for consistency by putting the exception in `.formatter.exs`.

apps/re/lib/accounts/user.ex
```diff
@@ -73,9 +76,21 @@ defmodule Re.User do
...
+    |> cast(params, @create_required ++ @create_optional)
+    |> validate_required(@create_required)
+    |> base_changeset()
+    |> Re.ChangesetHelper.generate_uuid()
+  end
+
   defp base_changeset(changeset) do
     changeset
     |> validate_email()
+    |> validate_phone()
```
rhnonose: Why do we need a phone validation? Isn't all users coming from account kit?

apps/re/lib/seller_leads/policy.ex
```diff
@@ -7,6 +7,10 @@ defmodule Re.SellerLeads.Policy do
 
   def authorize(_, %User{role: "admin"}, _), do: :ok
   def authorize(:create_seller_lead, %User{role: "user"}, _), do: :ok
+
+  def authorize(:create_broker_seller_lead, %User{role: "user", type: "partner_broker"}, _),
```
rhnonose: I'm not sure about mixing `type` with `role`.
`role` is a special field to handle authorization whereas `type` is a field to distinguish `user` roles.
I don't think we should restrict this permission by type.

apps/re/lib/seller_leads/seller_leads.ex
```diff
@@ -23,6 +29,37 @@ defmodule Re.SellerLeads do
...
+      name: Map.get(params, :owner_name),
+      phone: Map.get(params, :owner_telephone),
+      email: Map.get(params, :owner_email),
+      role: "user"
+    }
+
+    property_owner =
+      %User{}
+      |> User.create_changeset(property_owner_param)
+      |> handle_property_owner()
```
rhnonose: This shouldn't be a user since it doesn't go through `account_kit`.
There's a `owner_contact` table that you can use it for.
Creating a user here "manually" might result in some undesired behavior and security problems.

apps/re_web/lib/graphql/resolvers/seller_leads.ex
```diff
@@ -2,11 +2,30 @@ defmodule ReWeb.Resolvers.SellerLeads do
...
 
   def create_site(%{input: params}, %{context: %{current_user: current_user}}) do
     with :ok <- Bodyguard.permit(SellerLeads, :create_seller_lead, current_user, params) do
       SellerLeads.create_site(params)
     end
   end
+
+  def create_broker(%{input: params}, %{context: %{current_user: current_user}}) do
+    with :ok <- Bodyguard.permit(SellerLeads, :create_broker_seller_lead, current_user, params),
+         {:ok, address} <- get_address(params) do
```
rhnonose: You can call `Addresses.insert_or_update/1` directly here.

apps/re_web/lib/graphql/types/seller_lead.ex
```diff
@@ -24,12 +28,35 @@ defmodule ReWeb.Types.SellerLead do
...
+    field :owner_name, non_null(:string)
+    field :owner_telephone, non_null(:string)
+    field :owner_email, :string
+    field :address, non_null(:address_input)
+    field :campaign, :string
+    field :medium, :string
+    field :source, :string
+    field :initial_campaign, :string
+    field :initial_medium, :string
+    field :initial_source, :string
```
rhnonose: We should create an input object for UTM stuff for reuse and make it more readable.

apps/re_web/lib/graphql/types/seller_lead.ex
```diff
@@ -24,12 +28,35 @@ defmodule ReWeb.Types.SellerLead do
     field :price_request_id, non_null(:id)
   end
 
+  input_object :broker_seller_lead_input do
+    field :complement, :string
+    field :type, non_null(:string)
+    field :additional_information, :string
+    field :owner_name, non_null(:string)
+    field :owner_telephone, non_null(:string)
+    field :owner_email, :string
```
rhnonose: We should group owner contact in its own input object for better structure.

apps/re/priv/repo/migrations/20190730164027_create_broker_seller_leads.exs
```diff
@@ -0,0 +1,18 @@
...
+
+  def change do
+    create table(:broker_seller_leads) do
+      add :uuid, :uuid, primary_key: true
+      add :broker_uuid, references(:users, column: :uuid, type: :uuid)
+      add :address_uuid, references(:addresses, column: :uuid, type: :uuid)
+      add :owner_uuid, references(:owner_contacts, column: :uuid, type: :uuid)
+      add :complement, :string
+      add :type, :string
+      add :additional_information, :string
```
rhnonose: This should be of type `text` since it imply anything can go.

apps/re/lib/seller_leads/seller_leads.ex
```diff
@@ -23,6 +29,36 @@ defmodule Re.SellerLeads do
...
+    %Broker{}
+    |> Broker.changeset(attrs)
+    |> Repo.insert()
+  end
+
+  defp handle_property_owner(property_owner_changeset) do
+    phone = Changeset.get_field(property_owner_changeset, :phone)
+
+    case OwnerContacts.get_by_phone(phone) do
+      {:error, _} -> Repo.insert!(property_owner_changeset)
```
rhnonose: You can use `Repo.insert_or_update/1`.
Also, we don't generally use the `!` version of a function in general because we prefer handling an error tuple rather than an exception.
Exceptions (teehee) would be during `ecto_job`s where we have to raise an exception to be able to rerun a job.

apps/re/test/seller_leads/broker_lead_test.exs
```diff
@@ -0,0 +1,51 @@
...
+  describe "changeset" do
+    test "should be valid" do
+      user = insert(:user, type: "partner_broker")
+      owner = insert(:user)
+      address = insert(:address)
+
+      attr = @valid_attributes
+              |> Map.merge(%{broker_uuid: user.uuid})
+              |> Map.merge(%{address_uuid: address.uuid})
+              |> Map.merge(%{owner_uuid: owner.uuid})
```
rhnonose: You can use just one `Map.merge/2` function with a map with those 3 key/value pairs.

apps/re_web/lib/graphql/resolvers/seller_leads.ex
```diff
@@ -2,11 +2,27 @@ defmodule ReWeb.Resolvers.SellerLeads do
...
     end
   end
+
+  def create_broker(%{input: params}, %{context: %{current_user: current_user}}) do
+    with :ok <- Bodyguard.permit(SellerLeads, :create_broker_seller_lead, current_user, params),
+         {:ok, address} <- Addresses.insert_or_update(params.address) do
+      attr =
+        params
+        |> Map.merge(%{broker_uuid: current_user.uuid})
+        |> Map.merge(%{address_uuid: address.uuid})
```
rhnonose: Same thing as above, you can use only one `Map.merge/2` function.

apps/re_web/lib/graphql/types/seller_lead.ex
rhnonose: We should create a generic `utm_input` resource to reduce verbosity.

apps/re_web/test/graphql/seller_leads/mutation_test.exs
```diff
@@ -139,4 +144,151 @@ defmodule ReWeb.GraphQL.SellerLeads.MutationTest do
...
+          "type" => "Apartamento"
+        }
+      }
+      conn =
+        post(conn, "/graphql_api", AbsintheHelpers.mutation_wrapper(@create_mutation, variables))
+
+      assert %{"brokerSellerLeadCreate" =>  uuid} = json_response(conn, 200)["data"]
+    end
+
+    test "property owner user should not add broker lead", %{
```
rhnonose: This test should't be passing as you removed the broker type checking.
Did I miss something?

PR #694, opened 2019-08-01
Salesforce job queue to insert events

gabiseabra:


apps/re_integrations/lib/salesforce/client.ex
```diff
@@ -9,6 +9,9 @@ defmodule ReIntegrations.Salesforce.Client do
 
   @api_headers [{"Authorization", @api_key}, {"Content-Type", "application/json"}]
 
+  def insert(payload, :Event),
```
rhnonose: This shoulb be just a `insert_event/1` since we only have one insertion for now.
We can think about abstracting it later.

PR #696, opened 2019-08-02
Routific + salesforce integration

gabiseabra:
* Remove routific job queue and monitor jobs in salesforce's job queue
* Create salesforce events when the routific job finishes


pmargreff:
We pair reviewed it, I gave some minor feedback points. The general idea and implementation LGTM.

PR #697, opened 2019-08-02
Price suggestion query

rhnonose:
Add a query to be called by `garagem` so we separate from `requestPriceSuggestion`.
This allows us to use the new parameters to be consulted rather than just `suggestedPrice`.
We might use this to replace `requestPriceSuggestion` in the frontend since it comes with new parameters and we can have a separate mutation to create a new `price_suggestion_request` in the database.


apps/re_web/lib/graphql/resolvers/interests.ex
```diff
@@ -19,6 +23,12 @@ defmodule ReWeb.Resolvers.Interests do
     end
   end
 
+  def price_suggestion(%{input: params}, %{context: %{current_user: current_user}}) do
```
pmargreff: It isn't clear to me why it's growing inside `Interests` resolver. Shouldn't it have it own resolver now?

rhnonose: Yes, I thought about refactoring in this PR but it would get quite big, so I would be doing that later. Will create an improvement for that.

pmargreff: Roger that.

rhnonose:
@pmargreff The intention of this query is actually to separate the lead creating from the previous model to a pure query. It's restricted for admins for now since we'll still require a separate mutation to actually create a lead based on.

Having different queries/mutations for users/admins kind of breaks graphql's purpose, so I'm trying to find some middle ground.

PR #698, opened 2019-08-07
Salesforce update opportunity

gabiseabra:
Update salesforce opportunity stage to `visit_scheduled` after an event is inserted.


apps/re_integrations/lib/salesforce/job_queue.ex
```diff
@@ -33,9 +33,30 @@ defmodule ReIntegrations.Salesforce.JobQueue do
...
+    |> Repo.transaction()
+  end
+
+  def perform(%Multi{} = multi, %{
+        "type" => "update_opportunity",
+        "id" => id,
+        "opportunity" => payload
+      }) do
+    multi
+    |> Multi.run(:update_opportunity, fn _repo, _changes ->
```
pmargreff: you can use [`Multi.update` here](https://hexdocs.pm/ecto/Ecto.Multi.html#update/4). Multi.prun are made to use when you need changes already performed on past multi ops or when you need to use a function to rollback. When you only need `insert` and `update` the correspondent functions are good enough.

gabiseabra: `Salesforce.update_opportunity/2` sends a PUT request to `@emcasa/salesforce`'s api and returns the JSON response, it doesn't change the db.

PR #700, opened 2019-08-08
fix(migrations): adding explicit rollback for migrations

beligante:


apps/re/priv/repo/migrations/20190802180552_add_fields_to_calendars.exs
```diff
@@ -1,11 +1,19 @@
 defmodule Re.Repo.Migrations.AddFieldsToCalendars do
   use Ecto.Migration
 
-  def change do
+  def up do
     alter table(:calendars) do
       add(:name, :string)
       add(:speed, :string)
       modify(:external_id, :string, null: true)
```
pmargreff: You can add [`from` option on `modify`](https://hexdocs.pm/ecto_sql/Ecto.Migration.html#modify/3) here to make this migration reversible instead add a `down` function.

PR #703, opened 2019-08-09
Add check_duplicated function to check similar addresses

beligante:
This is the first part of this kind of checking. We are only checking duplicated leads against the seller leads schema.


apps/re/lib/addresses/address.ex
```diff
@@ -26,6 +26,7 @@ defmodule Re.Address do
 
     has_many :listings, Re.Listing
     has_many :developments, Re.Development
+    has_many(:seller_leads, Re.SellerLead, foreign_key: :address_uuid, references: :uuid)
```
rhnonose: Please remove the `()` for formatting consistency adding this as exception in `.formatter.exs`

PR #709, opened 2019-08-12
Switch lead creation to salesforce REST API

rhnonose:
To be able to save the lead ID into our database to perform an update on subsequent user interactions, I'm switching the zapier client call to our REST API call.

### DIFF
--- a/config/dev.secret-example.exs
+++ b/config/dev.secret-example.exs
@@ -35,7 +35,9 @@ config :re_integrations,
   orulo_client_token: "orulo_client_token",
   salesforce_event_owner_id: "salesforce_user_id",
   salesforce_api_key: "salesforce_api_key",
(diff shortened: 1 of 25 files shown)


config/dev.secret-example.exs
```diff
@@ -35,7 +35,9 @@ config :re_integrations,
   orulo_client_token: "orulo_client_token",
   salesforce_event_owner_id: "salesforce_user_id",
   salesforce_api_key: "salesforce_api_key",
-  salesforce_url: "salesforce_api_url"
+  salesforce_url: "salesforce_api_url",
+  salesforce_buyer_lead_record_id: "buyer_record_id",
```
pmargreff: This env var isn't used.

rhnonose: ayy

PR #718, opened 2019-08-14
Edit seller leads

rhnonose:
When a `site_seller_lead` comes in, it's the same lead as the `price_suggestion_request` that comes before.
To be able to unify that lead, we're saving some references and retrieving it to, instead of creating a new `seller_lead`, editing an existing one.
Then, when sending to salesforce, we're making an update instead of creating a new one and changing the `evaluation` parameter to be able to specify that this lead has "evolved".


rhnonose:
Yep @pmargreff 
It's a little hard to detach the logic because we rely so much on `multi` to reliably finish the job.
I'll take some time to refactor everything.

PR #721, opened 2019-08-14
Shortlist/get salesforce opportunity

pmargreff:
Create the module to receive a salesforce entity and parse it into usable params in the data service (basically `salesforce_opportunity` schema). 

---

I think you're good to go now, some of the next steps are
- plug in the shortlist service
- fetch listings returned by the service and expose them through GraphQL. 

This should be the very first iteration on shortlists, after that, we have a lot of steps to test, as save them and make it editable, return them to salesforce, so on...

### DIFF
--- a/apps/re/lib/shortlists/shortlists.ex
+++ b/apps/re/lib/shortlists/shortlists.ex
@@ -1,0 +1,24 @@
+defmodule Re.Shortlists do
+  @moduledoc """
+  Context for shortlists.
+  """
+
+  alias __MODULE__.{
+    Salesforce.Opportunity
+  }
+
+  alias Re.Salesforce
+
+  @behaviour Bodyguard.Policy
+
+  defdelegate authorize(action, user, params), to: Re.Shortlists.Policy
+
+  def generate_shortlist_from_salesforce_opportunity(opportunity_id) do
+    with {:ok, opportunity} <- Salesforce.get_opportunity(opportunity_id),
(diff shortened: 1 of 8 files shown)


pmargreff:
I think you're good to go now, some of the next steps are
- plug in the shortlist service
- fetch listings returned by the service and expose them through GraphQL. 

This should be the very first iteration on shortlists, after that, we have a lot of steps to test, as save them and make it editable, return them to salesforce, so on...

apps/re/lib/shortlists/shortlists.ex
```diff
@@ -0,0 +1,24 @@
...
+
+  alias Re.Salesforce
+
+  @behaviour Bodyguard.Policy
+
+  defdelegate authorize(action, user, params), to: Re.Shortlists.Policy
+
+  def generate_shortlist_from_salesforce_opportunity(opportunity_id) do
+    with {:ok, opportunity} <- Salesforce.get_opportunity(opportunity_id),
+         {:ok, service_params} <- Opportunity.build(opportunity) do
```
rhnonose: Since you're repeating this result in the `do` block, maybe it should be there altogether?
(I couldn't find what `validate/1` in `Opportunity` does, but it'll be better if `build/1` returns a clean ok/error tuple.

pmargreff: Yes, I'm thinking about it, at this moment I don't know which errors a wanna show to our clients and which ones I'll handle, I'll probably come back to this in a nearby future.

PR #726, opened 2019-08-15
opportunity route info

gabiseabra:
Update salesforce opportunity with `route_url` and `route_unserved_reason` after passing through routific's api.


apps/re_integrations/lib/salesforce/payload/opportunity.ex
```diff
@@ -85,7 +91,23 @@ defmodule ReIntegrations.Salesforce.Payload.Opportunity do
...
-  defp changeset(struct, params), do: cast(struct, params, @params)
+  defp changeset(struct, params) do
+    struct
+    |> cast(params, @params)
+    |> put_route_url()
+  end
+
+  defp route_url(id), do: "#{@routific_job_url}/#{id}"
+
+  defp put_route_url(changeset) do
```
pmargreff: you can rewrite it without the `case` statement, just pattern matching the changeset in the function params.

config/prod.exs
```diff
@@ -87,6 +87,7 @@ config :re_integrations,
   orulo_client_token: System.get_env("ORULO_CLIENT_TOKEN"),
   routific_url: System.get_env("ROUTIFIC_URL"),
   routific_api_key: System.get_env("ROUTIFIC_API_KEY"),
+  routific_job_url: System.get_env("ROUTIFIC_JOB_URL"),
```
pmargreff: could you add this var on `dev.secret` as well?

PR #727, opened 2019-08-16
add listings to the check of duplicity on seller leads creation

beligante:


apps/re/lib/seller_leads/seller_leads.ex
```diff
@@ -82,11 +82,19 @@ defmodule Re.SellerLeads do
...
+  defp check_attribute_duplicated(address, complement, attribute_to_fetch) do
     address
-    |> Repo.preload(:seller_leads)
-    |> Map.get(:seller_leads)
-    |> Enum.any?(fn seller_lead ->
-      normalize_complement(seller_lead.complement) == normalize_complement(complement)
+    |> Repo.preload(attribute_to_fetch)
+    |> Map.get(attribute_to_fetch)
+    |> Enum.any?(fn attribute ->
+      normalize_complement(attribute.complement) == complement
```
pmargreff: shouldn't you normalize complement as well?

apps/re/lib/seller_leads/seller_leads.ex
```diff
@@ -82,11 +82,19 @@ defmodule Re.SellerLeads do
   end
 
   def duplicated?(address, complement) do
+    normalized_complement = normalize_complement(complement)
+
+    address
```
pmargreff: You can rewrite it as `check_attribute_duplicated(address, normalized_complement, :seller_leads) || check_attribute_duplicated(address, normalized_complement, :listings)` it's not so elegant, but it's less tricker and easy to get what is happening here.

PR #729, opened 2019-08-16
Shortlist/service client

pmargreff:
Allow a client request non-persistent shortlists through GraphQL. 

Right now we're not persisting any shortlists, the backend receives an id from the client, request an opportunity to salesforce, parse it to shortlist expected params, and return some listings to the user.

The client is not done yet, so I'm awaiting it to mark the PR as ready to review.  

The next steps are to structure shortlist, liked and disliked listings as well find a trigger to create shortlists automatically. Probably I'll involve more people in both discussions...


apps/re_web/lib/graphql/resolvers/shortlists.ex
```diff
@@ -0,0 +1,13 @@
+defmodule ReWeb.Resolvers.Shortlists do
+  @moduledoc """
+  Resolver module for shortlists queries and mutations
+  """
+  alias Re.Shortlists
+
+  def index(%{opportunity_id: opportunity_id}, %{context: %{current_user: current_user}}) do
+    case Bodyguard.permit(Shortlists, :related_listings, current_user, %{}) do
```
rhnonose: We have a separate `related` listings, this might be confusing.

pmargreff: absolutely

PR #730, opened 2019-08-19
Get list of duplicated entities on seller lead

beligante:
This PR allow us to know what entities are duplicated for an seller lead address and complement. This data will be used during the integration with salesforce to make easier to track if the lead is really a duplicated one

### DIFF
--- a/apps/re/lib/seller_leads/seller_leads.ex
+++ b/apps/re/lib/seller_leads/seller_leads.ex
@@ -91,11 +91,14 @@ defmodule Re.SellerLeads do
   end
 
   def duplicated?(address, complement) do
-    duplicated_entities = duplicated_entities(address, complement)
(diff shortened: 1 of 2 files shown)


apps/re/lib/seller_leads/seller_leads.ex
```diff
@@ -96,16 +96,12 @@ defmodule Re.SellerLeads do
   end
 
   def duplicated?(duplicated_entities \\ %{}) do
-    duplicated_entities
-    |> Map.values()
-    |> Enum.any?(fn value -> Kernel.length(value) > 0 end)
+    Kernel.length(duplicated_entities) > 0
```
rhnonose: You can just pattern-match with `[]` in the function definition:
```
  def duplicated?([]), do: false
  def duplicated?(_), do: true
```

PR #732, opened 2019-08-19
Add duplicated flag to seller lead

beligante:
With this pull request we are saving the ids of the entities that are apparently duplicated in the database.

- [x] Add duplicated flag to during seller_lead creation
- [x] Add duplicated entities to  seller_lead creation
- [x] Write some tests


apps/re/lib/addresses/addresses.ex
```diff
@@ -32,6 +32,13 @@ defmodule Re.Addresses do
     end
   end
 
+  def get_by_uuid(uuid) do
+    case Repo.get_by!(Address, uuid: uuid) do
```
rhnonose: `Repo.get_by!/2` with `!` raises an exception when it doesn't find the record, so please use `get_by/2`.

apps/re/lib/seller_leads/duplicity_checker.ex
```diff
@@ -0,0 +1,80 @@
...
+    |> Map.get(entity_name)
+    |> Enum.filter(fn entity ->
+      normalize_complement(entity.complement) == normalized_complement
+    end)
+    |> Enum.map(fn entity ->
+      %{type: entity.__struct__, uuid: entity.uuid}
+    end)
+  end
+
+  @number_group_regex ~r/(\d)*/
```
pmargreff: Take the chance to move this logic to `Addresses`, so we can reuse it in other places.

beligante: You mean only the regex right? not the complement normalization

PR #736, opened 2019-08-20
Move seller leads out of interests

rhnonose:
A little refactoring to move some `SellerLeads` functions out of `Interests`.
Next refactorings:
- Rename `Interests` to appropriate naming
- Move it to `BueyrLeads`
- Rename in the API and deprecate old ones


apps/re_web/lib/graphql/types/dashboard.ex
```diff
@@ -47,9 +47,27 @@ defmodule ReWeb.Types.Dashboard do
...
+    field :listing_price_error_q90_max, :float
+    field :listing_price_per_sqr_meter, :float
+    field :listing_average_price_per_sqr_meter, :float
+  end
+
   object :dashboard_queries do
     @desc "Get dashboard stats"
     field :dashboard, :dashboard, resolve: &DashboardResolvers.index/2
+
+    @desc "Query price suggestion"
```
pmargreff: Is that right? The dashboard shouldn't be used only to follow states? What were the criteria to move it here? This one will be moved again?

rhnonose: Likely, yeah.
I created a query for price suggestions to not follow the same flow as a lead creation.
I'm still not sure if we should use this query and then create a lead in case we use the same logic in the frontend.
If we decide to follow that design, we should then move again to the appropriate location.

PR #739, opened 2019-08-21
add query on graphql for duplicity checking on seller_lead

beligante:
This query will be user on Parter Broker flux for lead creation to a faster feedback about the duplicity status during it's lead creation


apps/re_web/test/graphql/duplicity_checker/query_test.exs
```diff
@@ -0,0 +1,196 @@
...
+      }
+
+      conn =
+        post(
+          conn,
+          "/graphql_api",
+          AbsintheHelpers.query_wrapper(@check_duplicity_query, variables)
+        )
+
+      refute json_response(conn, 200)["data"]["checkDuplicity"]
```
rhnonose: You should specifically assert for a false return: `assert false == ...`
This is because if a query fails for some reason, it doesn't fill `"checkDuplicity"` and it fills `"errors"`, so this test won't break because `refute` passes with a `nil` return.

apps/re_web/test/graphql/duplicity_checker/query_test.exs
```diff
@@ -0,0 +1,196 @@
...
```
rhnonose: You should specifically assert a `Unauthenticated` error here.
Note that your tests are the same for allowed/not allowed.

apps/re_web/test/graphql/duplicity_checker/query_test.exs
```diff
@@ -0,0 +1,196 @@
...
```
rhnonose: Same as above.

apps/re_web/test/graphql/duplicity_checker/query_test.exs
```diff
@@ -0,0 +1,196 @@
...
+      }
+
+      conn =
+        post(
+          conn,
+          "/graphql_api",
+          AbsintheHelpers.query_wrapper(@check_duplicity_query, variables)
+        )
+
+      assert json_response(conn, 200)["data"]["checkDuplicity"]
```
rhnonose: You should specifically assert what's returned here.

apps/re_web/test/graphql/duplicity_checker/query_test.exs
```diff
@@ -0,0 +1,196 @@
...
+      }
+
+      conn =
+        post(
+          conn,
+          "/graphql_api",
+          AbsintheHelpers.query_wrapper(@check_duplicity_query, variables)
+        )
+
+      refute json_response(conn, 200)["data"]["checkDuplicity"]
```
rhnonose: Same as above.

apps/re_web/test/graphql/duplicity_checker/query_test.exs
```diff
@@ -0,0 +1,196 @@
...
+      }
+
+      conn =
+        post(
+          conn,
+          "/graphql_api",
+          AbsintheHelpers.query_wrapper(@check_duplicity_query, variables)
+        )
+
+      assert json_response(conn, 200)["data"]["checkDuplicity"]
```
rhnonose: Same as above.

PR #741, opened 2019-08-21
send duplicated information about seller lead to salesforce

beligante:


apps/re/lib/seller_leads/embedded/duplicated_entity.ex
```diff
@@ -18,6 +20,6 @@ defmodule Re.SellerLeads.DuplicatedEntity do
       Ecto.Changeset.cast(
         struct,
         params,
-        ~w(type uuid)a
+        ~w(type uuid)a@derive Jason.Encoder
```
pmargreff: is this right?

beligante: No it's not. Git is with "rosca"

PR #742, opened 2019-08-21
Shortlist/create shortlist structure

pmargreff:
This feature creates a proper structure to shortlists. A shortlists can have multiple listings, we have no trigger to create a new one, so the creation is made by `get_or_create`. If it doesn't exist the `get_or_create` will get listings from shortlist service and create a new shortlist from it, if it does exists will send the old one (as the name says). 

Is a good idea take a look on the base (`shortlist/service-client`) branch before digging into this one. After I create a `GraphQL` query for use this feature, I'll remove `generate_shortlist_from_salesforce_opportunity` feature, once this only sends data API results to the client instead of persist them. 

I'll implement GraphQL call in another PR.

### DIFF
--- a/apps/re/lib/shortlists/shortlists.ex
+++ b/apps/re/lib/shortlists/shortlists.ex
@@ -23,28 +23,28 @@ defmodule Re.Shortlists do
   def get_or_create(opportunity_id) do
     case Repo.get_by(Shortlist, opportunity_id: opportunity_id) do
       nil ->
-        with {:ok, opportunity} <- Salesforce.get_opportunity(opportunity_id),
-             {:ok, params} <- create_params(opportunity),
-             {:ok, listing_uuids} <- get_shortlist(params) do
-          listings = get_active_listings_by_uuid(listing_uuids)
-
-          %Shortlist{}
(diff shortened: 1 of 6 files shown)


apps/re/lib/shortlists/shortlists.ex
```diff
@@ -12,17 +12,39 @@ defmodule Re.Shortlists do
...
+  defp create_shortlist(opportunity_id) do
+    with {:ok, listing_uuids} <- get_listing_uuids_from_opportunity_preferences(opportunity_id) do
+      listings = get_active_listings_by_uuid(listing_uuids)
+
+      %Shortlist{}
+      |> Shortlist.changeset(%{opportunity_id: opportunity_id})
+      |> Ecto.Changeset.put_assoc(:listings, listings)
+      |> Repo.insert()
+    else
+      _error -> {:error, :invalid_opportunity}
```
rhnonose: This error handling should be inside `get_listing_uuids_from_opportunity_preferences/1` since it's related to getting `opportunity`.

pmargreff: The error should me generalist, so I change it, I have no sure which errors should or shouldn't be sent to clients.

PR #743, opened 2019-08-21
Visit schedule reports

gabiseabra:
Send visit schedule reports to slack via zapier.

### DIFF
--- a/apps/re_integrations/lib/salesforce/mapper/zapier.ex
+++ b/apps/re_integrations/lib/salesforce/mapper/zapier.ex
@@ -33,13 +33,14 @@ defmodule ReIntegrations.Salesforce.Mapper.Zapier do
 
   defp build_unserved(%{unserved: unserved}),
     do:
-      Enum.map(unserved, fn {id, reason} ->
+      unserved
+      |> Enum.map(fn {id, reason} ->
         "[<#{opportunity_url(id)}\">|#{id}>] #{reason}"
       end)
       |> Enum.join("\n")
 
   defp build_solution(%{solution: solution}),
-    do: Enum.map(solution, &build_route/1) |> Enum.join("\n")
+    do: solution |> Enum.map(&build_route/1) |> Enum.join("\n")
 
   defp build_route({calendar_uuid, [_depot | visits]}),
     do:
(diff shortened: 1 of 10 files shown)


apps/re_integrations/lib/salesforce/zapier_client.ex
```diff
@@ -0,0 +1,66 @@
+defmodule ReIntegrations.Salesforce.ZapierClient do
```
pmargreff: This module knows a lot, the client should only know how to interact with the external world. You can extract things like, `report` logic (mount the message) to the context module (and create isolated tests to it), the same can applied in `build_route` fn, you can extract and split into two different functions, interact with the client and another one only receives and parse the response, it can help to test .

apps/re_integrations/lib/salesforce/mapper/zapier.ex
```diff
@@ -0,0 +1,61 @@
+defmodule ReIntegrations.Salesforce.Mapper.Zapier do
```
pmargreff: A simple test case to this module would be great.

apps/re_integrations/test/salesforce/mapper/zapier_test.exs
```diff
@@ -0,0 +1,73 @@
...
+                 status: :finished,
+                 options: %{"date" => "2019-08-01T21:00:00.000+0000"},
+                 unserved: %{},
+                 solution: %{
+                   @calendar_uuid => [@visit]
+                 }
+               })
+
+      assert String.contains?(body, "Sessões de tour agendadas")
+      assert not String.contains?(body, "Opotunidades não agendadas")
```
pmargreff:
```suggestion
      refute String.contains?(body, "Opotunidades não agendadas")
```

apps/re_integrations/test/salesforce/mapper/zapier_test.exs
```diff
@@ -0,0 +1,73 @@
...
+               Mapper.Zapier.build_report(%Routific.Payload.Inbound{
+                 status: :finished,
+                 options: %{"date" => "2019-08-01T21:00:00.000+0000"},
+                 unserved: %{
+                   "0x01" => "unserved reason"
+                 },
+                 solution: %{}
+               })
+
+      assert not String.contains?(body, "Sessões de tour agendadas")
```
pmargreff:
```suggestion
      refute String.contains?(body, "Sessões de tour agendadas")
```

PR #748, opened 2019-08-23
Salesforce/validate opportunities

gabiseabra:
Validate required inputs from salesforce opportunities before sending to routific's api

### DIFF
--- a/apps/re_integrations/lib/salesforce/salesforce.ex
+++ b/apps/re_integrations/lib/salesforce/salesforce.ex
@@ -90,8 +90,6 @@ defmodule ReIntegrations.Salesforce do
     do: Enum.reduce(errors, multi, &update_failed_opportunity/2)
 
   defp update_failed_opportunity({:error, error, %{id: id}, %{errors: errors}}, multi) do
-    IO.inspect(errors)
-
     JobQueue.enqueue(multi, "update_#{id}", %{
       "type" => "update_opportunity",
       "id" => id,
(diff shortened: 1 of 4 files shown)


apps/re_integrations/lib/salesforce/salesforce.ex
```diff
@@ -28,13 +28,14 @@ defmodule ReIntegrations.Salesforce do
   end
 
   def update_opportunity(id, payload) do
-    with {:ok, opportunity} <- Opportunity.validate(payload),
+    with opportunity <- struct(%Opportunity{}, payload),
```
rhnonose: Using `struct` here won't run the validations and result in having invalid data.
It should go through a changeset.

apps/re_integrations/lib/salesforce/salesforce.ex
```diff
@@ -28,13 +28,14 @@ defmodule ReIntegrations.Salesforce do
...
          {:ok, %{status_code: 200, body: body}} <- Client.update_opportunity(id, opportunity),
          {:ok, data} <- Jason.decode(body) do
       {:ok, data}
     else
       {:ok, %{status_code: _status_code} = data} -> {:error, data}
-      error -> error
+      {:error, error} -> {:error, error}
+      error -> {:error, error}
```
rhnonose: This error handling is getting confusing here.
Can't you make sure downstream that errors are returned in the correct format and you don't have to handle it here?

PR #752, opened 2019-08-26
Build routific fleet breaks from google calendar events

gabiseabra:


gabiseabra:
> Any specific reason to be using `tesla`?
> Tesla is not a explicit dependency of this project.
> Unless there's a reason to use it over `httpoison`, we should avoid using different clients as it makes the code a little bit confusing.
> This PR overall needs a better approach.
> `Payload.Outbound` is doing more than it should with `get_breaks/2` function, which is also hard to understand what it's doing.
> Maybe you can pair with someone to split into bits and make it more readable.

Tesla is a dependency of `google_api_calendar`, it can't be configured. Should I list it in `mix.exs`?

rhnonose:
Then, it's ok to use it, no need to put in `mix.exs`.
Please add a comment in the test file that tesla is being used because of this limitation.

apps/re_integrations/lib/routific/payload/outbound/vehicle.ex
```diff
@@ -1,6 +1,10 @@
 defmodule ReIntegrations.Routific.Payload.Outbound.Vehicle do
   @moduledoc """
   Module for mapping calendars to routific vehicles.
+  A vehicle is an object in routific's `fleet` parameter,
+  it represents a driver which may serve a route.
```
rhnonose: ~which~ whom

apps/re_integrations/lib/routific/payload/outbound/vehicle.ex
```diff
@@ -34,34 +38,40 @@ defmodule ReIntegrations.Routific.Payload.Outbound.Vehicle do
 
   defp get_breaks(%{external_id: external_id} = calendar, opts) when is_binary(external_id) do
```
pmargreff: Does this name still valid? Something like `get_filled_entries`, or `get_unavailable_entries` could have a strong meaning?

gabiseabra: I think so, since this is the terminology for "unavailable time windows" in routific's context. There's a more specific underlying function (`get_unavailable_times`) to get events from the photographer's calendars in this module.

PR #755, opened 2019-08-28
Attribute address to params to use normalized postal_code

rhnonose:
`SellerLeads` was not passing the inserted address to price suggestions, just the id to save it, so it was going with the address's postal code without normalizing it.


apps/re/test/seller_leads/seller_leads_test.exs
```diff
@@ -215,6 +215,47 @@ defmodule Re.SellerLeadsTest do
...
+        scheme: "http"
+      }
+
+      body =
+        ~s({"payload":{"zip_code":"11111000","type":"APARTMENT","suites":1,"street_number":"100","parking":1,"lng":-10.101,"lat":-10.101,"condo_fee":300,"bedrooms":2,"bathrooms":1,"area":80}})
+
+      assert_called(HTTPoison, :post, [
+        ^uri,
+        ^body,
+        [{"Content-Type", "application/json"}, {"X-Api-Key", "mahtoken"}]
```
pmargreff: Could use env var value here?

PR #757, opened 2019-08-29
Salesforce/description

gabiseabra:
This adds `set_up_pinkbox`, `key_location`, and the opportunity's url to the description of events inserted from the tour scheduling flow.


apps/re_integrations/lib/salesforce/mapper/routific.ex
```diff
@@ -61,7 +66,10 @@ defmodule ReIntegrations.Salesforce.Mapper.Routific do
          "SDR: #{sdr_name}\n" <>
            "Cliente: #{account_name}\n" <>
            "Telefone: #{account_phone}\n" <>
-           if(is_nil(event.notes), do: "", else: event.notes)
+           "Instalar PinkBox: #{format_bool(event.custom_notes["set_up_pinkbox"])}\n" <>
+           "Localização da chave: #{format_key_location(event.custom_notes["key_location"])}\n\n" <>
```
rhnonose: You can use the `||` operator inline: `event.custom_notes["key_location"] || "Não informado"`.
You can also refactor the next line to use the same logic instead of using an `if`.

apps/re_integrations/test/salesforce/mapper/routific_test.exs
```diff
@@ -20,7 +20,12 @@ defmodule ReIntegrations.Salesforce.Mapper.RoutificTest do
     break: false,
     idle_time: 0,
     notes: "",
-    custom_notes: %{"account_id" => "0x01", "owner_id" => "0x01"}
+    custom_notes: %{
+      "account_id" => "0x01",
+      "owner_id" => "0x01",
+      "set_up_pinkbox" => true,
+      "key_location" => "x"
```
pmargreff: could you create a test against `nil` value?

apps/re_integrations/lib/salesforce/mapper/routific.ex
pmargreff: You can also remove if inline in favor to use functions, those can grow up in the wrong place and become hard to read really fast.

apps/re_integrations/test/salesforce/mapper/routific_test.exs
```diff
@@ -148,5 +132,35 @@ defmodule ReIntegrations.Salesforce.Mapper.RoutificTest do
...
+          custom_notes: %{
+            "account_id" => "0x01",
+            "owner_id" => "0x01"
+          }
+        })
+        |> Mapper.Routific.build_event(@calendar_uuid, @payload)
+
+      assert String.contains?(description, "chave: Não informado")
+      assert String.contains?(description, "Pinkbox: Não")
+    end
```
rhnonose: These should be 2 different test cases.

PR #758, opened 2019-08-30
feat(salesforce) adjust visit time windows to avoid lateness

gabiseabra:


apps/re_integrations/lib/salesforce/payload/opportunity.ex
```diff
@@ -60,7 +60,11 @@ defmodule ReIntegrations.Salesforce.Payload.Opportunity do
   @required ~w(account_id owner_id address city tour_period)a
 
   @tour_visit_duration Application.get_env(:re_integrations, :tour_visit_duration, 40)
-  @tour_visit_max_lateness Application.get_env(:re_integrations, :tour_visit_max_lateness, 10)
+  @tour_visit_max_antecipation Application.get_env(
```
rhnonose: Should be `anticipation` 🙂 .

PR #759, opened 2019-08-30
Fix suggested_price resolver

gabiseabra:
This fixes listings' `suggested_price` field resolving to a price teller response object when it's value is null in the database.


apps/re_web/test/graphql/listings/query_test.exs
```diff
@@ -1436,6 +1436,37 @@ defmodule ReWeb.GraphQL.Listings.QueryTest do
...
+          listing (id: $id) {
+            id
+            suggested_price
+          }
+        }
+      """
+
+      conn = post(conn, "/graphql_api", AbsintheHelpers.query_wrapper(query, variables))
+
+      assert %{"suggested_price" => 26279.0} = json_response(conn, 200)["data"]["listing"]
```
rhnonose: Please also assert that priceteller was called with `assert_called`.

PR #764, opened 2019-09-03
Add price suggestion jsonb field to listings

gabiseabra:
Store all properties from price suggestions to the listings table.


apps/re/lib/listings/schemas/listing.ex
```diff
@@ -45,6 +48,12 @@ defmodule Re.Listing do
     field :sold_price, :integer
     field :liquidity_ratio, :float
 
+    embeds_one(
+      :price_suggestion,
+      Re.PriceTeller.Output,
+      on_replace: :update
+    )
```
rhnonose: I'm not sure about reusing `Re.PriceTeller.Output`. A change in the API might break previously stored values if there's no migration.
Thoughts @pmargreff ?

apps/re_web/lib/graphql/resolvers/listings.ex
```diff
@@ -217,21 +218,28 @@ defmodule ReWeb.Resolvers.Listings do
...
-  def suggested_price(listing, _, %{context: %{current_user: current_user}}) do
-    case Bodyguard.permit(Listings, :suggested_price, current_user, listing) do
-      :ok -> do_suggest_price(listing)
-      _ -> {:ok, nil}
+  def price_suggestion(listing, _, %{context: %{current_user: current_user}}) do
+    with :ok <- Bodyguard.permit(Listings, :price_suggestion, current_user, listing),
+         {:ok, suggested_price} <- do_suggest_price(listing) do
+      {:ok, suggested_price}
+    else
+      _ -> {:ok, %PriceTeller.Output{}}
```
rhnonose: I'm not sure about returning an empty schema.
There's no need for an `else` clauase here since `do_suggest_price/1` takes care of that.

apps/re_web/lib/graphql/resolvers/listings.ex
```diff
@@ -217,21 +218,28 @@ defmodule ReWeb.Resolvers.Listings do
     end
   end
 
-  def suggested_price(listing, _, %{context: %{current_user: current_user}}) do
-    case Bodyguard.permit(Listings, :suggested_price, current_user, listing) do
-      :ok -> do_suggest_price(listing)
-      _ -> {:ok, nil}
+  def price_suggestion(listing, _, %{context: %{current_user: current_user}}) do
+    with :ok <- Bodyguard.permit(Listings, :price_suggestion, current_user, listing),
+         {:ok, suggested_price} <- do_suggest_price(listing) do
```
rhnonose: This second `with` clause is not necessary, this clause could be in the `do` body.

apps/re_web/lib/graphql/resolvers/listings.ex
```diff
@@ -217,21 +218,28 @@ defmodule ReWeb.Resolvers.Listings do
...
-  defp do_suggest_price(%{suggested_price: nil} = listing) do
-    case PriceSuggestions.suggest_price(listing) do
-      {:ok, suggested_price} -> {:ok, suggested_price.listing_price_rounded}
-      _error -> {:ok, nil}
+  def suggested_price(listing, _, %{context: %{current_user: current_user}}) do
+    with :ok <- Bodyguard.permit(Listings, :suggested_price, current_user, listing),
+         {:ok, suggested_price} <- do_suggest_price(listing) do
+      {:ok, suggested_price.listing_price_rounded}
+    else
+      _ -> {:ok, nil}
```
rhnonose: Same comments as the function before.

pmargreff:
Our listing structure is will be to have with it. Add it as an association or expose it to through a proper GraphQL`s resolver isn't an option?

rhnonose:
> Our listing structure is will be to heavy with it. Add it as an association or expose it to through a proper GraphQL`s resolver isn't an option?

Logically, I'm not sure.

This entire field is heavily dependant on listing's attributes and should be updated on listing changes. Keeping it in a different table may give the wrong impression that it's independent of the listing when it's not.

Also, I don't see any use in keeping in a different table since it's a 1 on 1 association and no query needs for that information.

I suggested an embedded schema since, logically, all those fields should be inside a listing individually, but that would be very verbose and those fields individually don't make sense (and there's no structural grouping better than a embedded schema if we don't need to query them).

Does it make sense?

pmargreff:
- I agree with your first argument, but the other two seems weak to me (if we follow the second strictly we shouldn't create any 1 ~ 1 assoc). 

IMO we have two entities, one is listing, other is price suggestion. They do have a strong relation, but the're not the same thing.
 
My main concern is, today we are making it in your monolith, once it's decoupled from the main repo, we only need to rewrite the `resolver` (if it's not to coupled). Also, inserting it on listings we need to be ok with the coupling we can ~will~ create while adding new features.  

All those points are food for thought, if you still not convinced about it, feel free to use this approach, if you think it's a reality but not the priority, keep it and create a new tech debt...

rhnonose:
I'm convinced that either one is good.
It's not a hard logic to change later, but also the resolver argument for switching is the best one.
Feel free to choose @gabiseabra 🙂

gabiseabra:
I agree those are two separate entities and decoupling them would be ideal, 
but I don't see many advantages in having a separate table for price suggestions over a json field right now.
This is straightforward, easy to replace, and it suits our need which is only to serve this data associated to a listing. 
I'll create an task in notion so we can take up this issue again in the future.

apps/re/lib/listings/schemas/listing.ex
rhnonose: Yeah.
If there's an API change, we have to change `Output`, but we won't have the new values stored because we would need to sync with the API again.

rhnonose:
> Could we remove suggestedPrice and point its resolver to price_suggestion equivalent field or mark it as deprecated? Have the same information in different places can lead people into weird situations.

I think it was done in another PR.

apps/re_web/lib/graphql/resolvers/listings.ex
rhnonose: I don't think answering an empty `Output` is a good error handling, but yeah, the `with` part is fine like this.

PR #766, opened 2019-09-04
Shortlist/integrate with selekta

pmargreff:
I was working against a mocked API, some changes are needed to work with the real API, changes are: 
- Fix input names.
- Added the authentication token in the header. 
- Changed the request method from `get` to `post`. 
- A proper encoder for `Re.Shortlists.Salesforce.Opportunity`.


apps/re/lib/shortlists/salesforce.opportunity.ex
```diff
@@ -44,10 +44,10 @@ defmodule Re.Shortlists.Salesforce.Opportunity do
     field :min_bathrooms, :integer
     field :min_garage_spots, :integer
     field :min_area, :integer
-    field :preference_floor, PreferenceFloor
-    field :elevators, :boolean
+    field :floor, Floor
+    field :has_elevators, :boolean
     field :nearby_subway, :boolean
-    field :neighborhoods, {:array, :string}
+    field :neighborhood, {:array, :string}
```
rhnonose: Why the renaming?
It's an array of neighborhoods, this might get confusing later.

pmargreff: 'cause the client has changed, @andresionek91 can you change it on selekta?

andresionek91: @pmargreff I'll change that

PR #767, opened 2019-09-04
add owner contact to seller lead

beligante:


apps/re/priv/repo/migrations/20190904134216_add_owner_contact_to_seller_lead.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Re.Repo.Migrations.AddOwnerContactToSellerLead do
+  use Ecto.Migration
+
+  def up do
+    alter table(:seller_leads) do
+      add :owner_uuid, references(:owner_contacts, column: :uuid, type: :uuid)
+    end
```
rhnonose: Please create an index for this new reference.

apps/re/priv/repo/migrations/20190904134216_add_owner_contact_to_seller_lead.exs
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def down do
```
rhnonose: Creating a new field doesn't require `up` and `down`, only `change`.

PR #768, opened 2019-09-04
Listing/remove suggested price

gabiseabra:


apps/re/lib/price_suggestions/price_suggestions.ex
```diff
@@ -35,10 +35,7 @@ defmodule Re.PriceSuggestions do
 
   defp persist_suggested_price({:ok, suggested_price}, listing) do
     listing
-    |> Listing.changeset(%{
-      price_suggestion: suggested_price,
-      suggested_price: suggested_price.listing_price_rounded
-    })
+    |> Listing.changeset(%{price_suggestion: suggested_price})
```
rhnonose: You can now rename this variable to match the new field, otherwise it might be confusing.

apps/re/lib/price_suggestions/price_suggestions.ex
```diff
@@ -97,12 +94,15 @@ defmodule Re.PriceSuggestions do
     end
   end
 
+  defp set_suggested_price(%{data: %Listing{}} = changeset, {:ok, suggested_price}) do
+    changeset
+    |> Request.changeset(%{price_suggestion: suggested_price})
```
rhnonose: Same rename suggested before.

apps/re/lib/price_suggestions/price_suggestions.ex
```diff
@@ -97,12 +94,15 @@ defmodule Re.PriceSuggestions do
     end
   end
 
+  defp set_suggested_price(%{data: %Listing{}} = changeset, {:ok, suggested_price}) do
```
rhnonose: Is any other scenario that this changeset is not from a `Listing`?
This seems too defensive to me.

gabiseabra: `Re.SellerLead` also uses this function to set `suggested_price`.

gabiseabra:
`suggested_price` is still present in the listing graphql type, but it's resolved to `price_suggestion.listing_price_rounded` since the base PR (#764).

apps/re/lib/price_suggestions/price_suggestions.ex
rhnonose: Hm, might be worth refactoring later.
You can leave like this in this PR.
Can you create an improvement?

PR #769, opened 2019-09-04
Process broker lead to seller lead job

beligante:


apps/re/test/seller_leads/brokers/brokers_test.exs
```diff
@@ -0,0 +1,109 @@
...
+    SellerLeads.Brokers,
+    SellerLeads.JobQueue,
+    OwnerContacts
+  }
+
+  alias Ecto.Multi
+
+  describe "create_broker_lead" do
+    test "should create owner as owner contact when it doesn't exists" do
+      assert {:error, :not_found} == OwnerContacts.get_by_phone("+5599999999999")
```
rhnonose: This assertion doesn't make sense.
Imagine if your test starts with this line breaking.
An assertion break means a test is broken, but this line breaking means the test is invalid, which is different.
If you don't have the confidence that your test case will be clean, your test suit is not reliable, which is a different concern.
So, this line shouldn't exist.

apps/re/test/seller_leads/brokers/brokers_test.exs
```diff
@@ -0,0 +1,109 @@
...
+    @tag dev: true
+    test "should raise a exception when the broker doesn't have a salesforce_id" do
+      user = insert(:user, salesforce_id: nil)
+      broker_lead = build(:broker, broker: user)
+               |> with_owner()
+               |> with_address()
+               |> insert()
+
+      assert_raise RuntimeError, fn ->
+        Brokers.process_broker_request(Multi.new(), broker_lead.uuid)
```
rhnonose: This function call should not raise an error.
We should raise an error in the `JobQueue` context since most of our error handling is done through error tuples.
You can use this test case to refactor it (and create a test case with a raise calling `JobQueue.perform/2`).

apps/re/test/support/factory.ex
```diff
@@ -422,4 +422,26 @@ defmodule Re.Factory do
...
+  def broker_factory do
+    %Re.SellerLeads.Broker{
+      uuid: UUID.uuid4(),
+      complement: Address.secondary_address(),
+      type: random(:listing_type)
+    }
+  end
+
+  def with_owner(struct) do
+    owner = insert(:owner_contact)
```
rhnonose: Maybe these functions should use `build` as well.
Thoughts @pmargreff ?

apps/re/test/seller_leads/brokers/brokers_test.exs
beligante: Oks! It makes sense

apps/re/test/seller_leads/brokers/brokers_test.exs
beligante: Got it

apps/re/test/support/factory.ex
beligante: It makes any difference being `build` instead of `insert`?

rhnonose: It gets inserted just once in the database, so just a little performance.
But it's also a little conceptual on fixtures.
@pmargreff started with this style of building it, so he may have something to say.

beligante: Talking with @pmargreff he was using build before insert because he believed the insert before another insert does not creates the right association between the entities. He asked made a small P.O.C and we proved that his believe was wrong. The association is created normally (same as using build). 

But still for performance, we agree to use `build` to have all inserts in the same transaction

apps/re/lib/seller_leads/seller_leads.ex
```diff
@@ -75,25 +61,6 @@ defmodule Re.SellerLeads do
     |> PubSub.publish_new("notify_when_covered")
   end
 
-  defp handle_create_broker(owner, params) do
```
pmargreff: Could you try to not refactor while writing new features? It's hard to get what you're only moving and what you're adding. It's a better idea to open other PR just moving code.

beligante: Ok! Will do that next time. Tks for the comment

apps/re/lib/seller_leads/brokers/brokers.ex
```diff
@@ -0,0 +1,100 @@
...
+
+  alias Ecto.{
+    Changeset,
+    Multi
+  }
+
+  def process_broker_request(%Multi{} = multi, uuid) do
+    Broker
+    |> Ecto.Query.preload([:address, :owner, :broker])
+    |> Repo.get_by(uuid: uuid)
```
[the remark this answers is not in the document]
beligante: K

PR #780, opened 2019-09-06
Add price field to broker leads

beligante:
Price field was added in the UI for Partner Broker Leads. This basically add a price field to the broker lead schema and also make it insertable throughout GraphQL API


apps/re_web/lib/graphql/types/seller_lead.ex
```diff
@@ -37,6 +37,7 @@ defmodule ReWeb.Types.SellerLead do
     field :owner, non_null(:owner_contact_input)
     field :address, non_null(:address_input)
     field :utm, :input_utm
+    field :price, :integer
```
rdiego26: O que acha de deixar esse campo não obrigatório para não quebrar a interface atual?

beligante: Got it! Tks!

PR #785, opened 2019-09-09
Feature/correlation seller UUID

xlDoug:
Added correlation uuid to the seller flow


beligante:
Why do we need this new column?

apps/re/lib/changeset_helper.ex
```diff
@@ -10,4 +10,13 @@ defmodule Re.ChangesetHelper do
   end
 
   def generate_uuid(changeset), do: Changeset.unique_constraint(changeset, :uuid, name: :uuid)
+
+  def generate_correlation_uuid(%{data: %{correlation_uuid: nil}} = changeset) do
```
beligante: It seems to me that this correlation logic it's something related only with the scope of the `seller lead` schema, are you intend to use this in some other scopes in the future?

xlDoug: This column will be responsible for tying all the ids of all our systems. Will correlate garage ids, SalesForce, etc. This correlation id will work only in the sales flow.

xlDoug:
Closed due to we decided to improve the process, creating a new entity called "SellerJourney" who is responsible to correlate all the seller flow

PR #787, opened 2019-09-09
Add account existency check at salesforce before it's creation to avoid duplicity

beligante:
Since we have some accounts already recorded at salesforce, we need to check if for a given telephone number there are an respective account with it at salesforce. With this check we avoid duplicated accounts there


apps/re/test/partner_brokers/job_queue_test.exs
```diff
@@ -11,15 +12,75 @@ defmodule Re.PartnerBrokers.JobQueueTest do
...
+    @uri_query %URI{path: "/api/v1/query"}
+
+    test "should create account when it doesn't exists at salesforce" do
+      mock(HTTPoison, [post: 3], fn
+        @uri_query, _, _ ->
+          @empty_response
+
+        @uri_account, _, _ ->
+          {:ok, %{status_code: 200, body: ~s({"success": true, "id": "123", "errors": []})}}
+      end)
```
rhnonose: You're adding a mock with 2 calls but asserting only one.
Is both being called? Then, both should be asserted.

apps/re/test/partner_brokers/job_queue_test.exs
```diff
@@ -54,14 +115,122 @@ defmodule Re.PartnerBrokers.JobQueueTest do
...
         ^uri,
         ^body,
         [{"Authorization", ""}, {"Content-Type", "application/json"}]
       ])
     end
+
+    test "should not create account when it already exists at salesforce" do
+      mock(HTTPoison, [post: 3], fn
+        @uri_query, _, _ -> @single_response
+      end)
```
rhnonose: This syntax is for declaring multiple mocks for the same module/function.
You could write a shorter version like this:
`mock(HTTPoison, :post, @single_response)`
(double-check with other examples).

apps/re/test/partner_brokers/job_queue_test.exs
```diff
@@ -54,14 +115,122 @@ defmodule Re.PartnerBrokers.JobQueueTest do
...
+        ^body,
+        [{"Authorization", ""}, {"Content-Type", "application/json"}]
+      ])
+    end
+
+    @tag capture_log: true
+    test "should consider the first salesforceid when there are multiple results for the same account phone" do
+      mock(HTTPoison, [post: 3], fn
+        @uri_query, _, _ -> @multiple_response
+      end)
```
rhnonose: Same as previous test.

apps/re/test/partner_brokers/job_queue_test.exs
```diff
@@ -54,14 +115,122 @@ defmodule Re.PartnerBrokers.JobQueueTest do
...
+        ^body,
+        [{"Authorization", ""}, {"Content-Type", "application/json"}]
+      ])
+    end
+
+    @tag capture_log: true
+    test "should log that an specific id at salesforce have duplicated phone number" do
+      mock(HTTPoison, [post: 3], fn
+        @uri_query, _, _ -> @multiple_response
+      end)
```
rhnonose: Same as previous test.

apps/re/lib/partner_brokers/partner_brokers.ex
```diff
@@ -42,7 +42,21 @@ defmodule Re.PartnerBrokers do
 
   def create_account(uuid) do
     with {:ok, user} <- Users.get_preloaded(uuid, :districts),
-         {:ok, %{"id" => id, "success" => true}} <- Salesforce.create_account(user),
+         do: search_account_at_salesforce(user)
+  end
+
+  defp search_account_at_salesforce(%{phone: phone} = user) do
```
pmargreff: `search_account_by_phone` is an option

PR #789, opened 2019-09-09
Fix liquidity range

priscilaportela:


apps/re/test/listings/liquidity_test.exs
```diff
@@ -66,7 +66,7 @@ defmodule Re.Listings.LiquidityTest do
     end
 
     test "normalize liquidity ratio beetween -0.15 and 0.15 to 5" do
-      assert 5 == Liquidity.normalize_liquidity_ratio(-0.15)
+      assert 4 == Liquidity.normalize_liquidity_ratio(-0.15)
```
pmargreff: This assertion should be moved to the above block.

